PR #2, opened 2023-03-21
Packit: initial enablement

lsm5:
This commit adds Packit configuration files which will trigger builds on
copr:`rhcontainerbot/packit-builds` on every PR and on
copr:`rhcontainerbot/podman-next` on every commit to main branch.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

### DIFF
--- a/qm-selinux.spec.rpkg
+++ b/qm-selinux.spec.rpkg
@@ -1,0 +1,93 @@
+# For automatic rebuilds in COPR
+
+# The following tag is to get correct syntax highlighting for this file in vim text editor
+# vim: syntax=spec
+
+%global debug_package %{nil}
+
+# container-selinux stuff (prefix with ds_ for version/release etc.)
+# Some bits borrowed from the openstack-selinux package
+%global selinuxtype targeted
+%global moduletype services
+%global modulenames qm
+
+%global _installscriptdir %{_prefix}/lib/qm
+
+# Usage: _format var format
+# Expand 'modulenames' into various formats as needed
+# Format must contain '$x' somewhere to do anything useful
+%global _format() export %1=""; for x in %{modulenames}; do %1+=%2; %%1+=" "; done;
+
+Name: {{{ git_dir_name }}}
+Epoch: 101
+Version: {{{ git_dir_version }}}
+Release: 1%{?dist}
+License: GPLv2
+URL: https://github.com/containers/container-selinux
+Summary: SELinux policies for container runtimes
+VCS: {{{ git_dir_vcs }}}
+Source: {{{ git_dir_pack }}}
+BuildArch: noarch
+BuildRequires: container-selinux
+BuildRequires: make
+BuildRequires: git-core
+BuildRequires: pkgconfig(systemd)
+BuildRequires: selinux-policy >= %_selinux_policy_version
+BuildRequires: selinux-policy-devel >= %_selinux_policy_version
+Requires: selinux-policy >= %_selinux_policy_version
+Requires(post): selinux-policy-base >= %_selinux_policy_version
+Requires(post): selinux-policy-targeted >= %_selinux_policy_version
+Requires(post): policycoreutils
+Requires(post): libselinux-utils
+
+%description
+SELinux policy modules for use with container runtimes.
+
+%prep
+{{{ git_dir_setup_macro }}}
+sed -i 's/install: man/install:/' Makefile
+
+# Remove unavailable tokens
+%if 0%{?fedora} <= 37 || 0%{?rhel} <= 9
+sed -i '/user_namespace/d' qm.if
+%endif
+
+%build
+%{__make}
+
+%install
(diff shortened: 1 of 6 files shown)


lsm5:
@rhatdan looks like `make man` can't be done without superuser.  Are you cool with adding a generated `qm_selinux.8` into the repo as in container-selinux ?

```
$ make man
make -f /usr/share/selinux/devel/Makefile qm.pp
make[1]: Entering directory '/home/lsm5/repositories/qm-selinux'
qm.if:14: Warning: duplicate definition of qm_domain_template(). Original definition on /usr/share/selinux/devel/include/services/qm.if:14.
Compiling targeted qm module
Creating targeted qm.pp policy package
rm tmp/qm.mod tmp/qm.mod.fc
make[1]: Leaving directory '/home/lsm5/repositories/qm-selinux'
Compressing qm.pp -> qm.pp.bz2
bzip2 -f -9 qm.pp
semodule -i qm.pp.bz2
libsemanage.semanage_create_store: Could not read from module store, active modules subdirectory at /var/lib/selinux/targeted/active/modules. (Permission denied).
```

qm_selinux.8
```diff
@@ -0,0 +1,345 @@
+.TH  "qm_selinux"  "8"  "23-03-21" "qm" "SELinux Policy qm"
```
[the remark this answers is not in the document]
lsm5: I've committed this because I'm unable to generate it in rpmbuild as it needs sudo access. Let me know if there's a way around it.

Also, the container-selinux repo does have `container-selinux.8` committed into the repo itself.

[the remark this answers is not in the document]
rhatdan: Ok, you are right. I am thinking this was like podman man pages, but this man page is generated from the policy.

qm-selinux.spec.rpkg
```diff
@@ -0,0 +1,99 @@
...
+%prep
+{{{ git_dir_setup_macro }}}
+sed -i 's/install: man/install:/' Makefile
+
+# Remove unavailable tokens
+%if 0%{?fedora} <= 37 || 0%{?rhel} <= 9
+sed -i '/user_namespace/d' qm.if
+%endif
+
+# selinux-policy on RHEL 8 isn't new enough
```
lsm5, line 53: @rhatdan looks like RHEL 8 selinux-policy is missing way too many things atm.

qm-selinux.spec.rpkg
```diff
@@ -0,0 +1,99 @@
...
+%doc README.md
+%dir %{_datadir}/selinux
+%{_datadir}/selinux/*
+%{_mandir}/man8/*
+%dir %{_sharedstatedir}/qm
+%dir %{_sharedstatedir}/qm/rootfs%{_datadir}
+%dir %{_sharedstatedir}/qm/rootfs%{_datadir}/containers
+%{_sharedstatedir}/qm/rootfs%{_datadir}/containers/container_contexts
+%dir %{_libexecdir}/%{name}
+%{_libexecdir}/%{name}/install.sh
```
lsm5, line 96: @rhatdan added the install.sh script to `/usr/libexec/qm-selinux` . Please verify the change and remind me what else I need to add.

qm-selinux.spec.rpkg
lsm5: ack, this can get simplified further then. Thanks!

qm-selinux.spec.rpkg
lsm5: > I think we should name this /usr/libexec/qm-install 

I think the script should be 1 level further down. So, inside `/usr/libexec/qm/` or if it's only to be used in conjunction with podman, maybe in `/usr/libexec/podman/`.

> I am thinking of renaming this repo to just qm

Dunno, I guess having the `-selinux` in the name provides a better idea, but not a stong opinion. We could check with the team if you prefer.

lsm5: > Yes, /usr/libexec/qm/install would be good. Or /usr/lib/qm/install since this is not to be executed by other users of a tool.
> 
> As far as the -selinux drop is concerned, the goal of this repo is setting up a QM isolated environment. The SELinux component is just a small part of it.

SGTM

qm-selinux.spec.rpkg
lsm5: centos 8 stream disabled.

rhatdan:
Ok I the setup script needs to be modified to expect that the qm-selinux package was already built.

lsm5:
> Ok I the setup script needs to be modified to expect that the qm-selinux package was already built.

@rhatdan ping, just checking if you've updated the script.

rhatdan:
I have not completed it yet, hopefully I will get back to this tomorrow.

rhatdan:
Ok I am going to merge, and then we can continue working and improving this.

PR #4, opened 2023-04-11
Fixup packet

rhatdan:


qm.spec.rpkg
```diff
@@ -58,7 +58,7 @@ sed -i '/user_namespace/d' qm.if
 %install
 # install policy modules
 %_format MODULES $x.pp.bz2
-%{__make} DESTDIR=%{buildroot} DATADIR=%{_datadir} install install-setup install-quadlet
+%{__make} DESTDIR=%{buildroot} DATADIR=%{_datadir} install
```
lsm5, line 61: We could remove the `DATADIR=%{_datadir}` as well since the defaults are the same as that expected by Fedora. But no biggie either way.

lsm5 (approved):
LGTM

PR #13, opened 2023-04-12
Packit: Use all available targets in copr

lsm5:
Looks like the `targets` key prevents packit from triggering copr builds on commit to main. This commit removes it so packit will use all available targets on the `rhcontainerbot/packit-builds` copr.

The only additional targets enabled would be aarch64 ones, which shouldn't take longer than the current time for packit builds.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,13 +16,6 @@ jobs:
     owner: rhcontainerbot
     project: packit-builds
     enable_net: true
-    # x86_64 is assumed by default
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

lsm5:
ugh, hmm centos-stream-8 gonna be trouble. I guess I'll just specify targets in both sections

lsm5:
cancelling in favor of https://github.com/containers/qm/pull/20

PR #14, opened 2023-04-12
fixup setup to run first time and not fail if qm.service is not enabled

rhatdan:


rhatdan:
@lsm5 @alexlarsson PTAL

setup
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -ex
+#! /bin/sh -ex
```
lsm5, line 1: just double checking this was intentional.

setup
```diff
@@ -15,16 +15,13 @@ INSTALLDIR="$1"
 ROOTFS="$2"
 [ ! -z "${ROOTFS}" ] || ROOTFS=/usr/lib/qm/rootfs
 
-. /etc/os-release
-
-systemctl stop qm 2> /dev/null
+systemctl stop qm.service 2>/dev/null || true
 
 install() {
     rootfs=$1
```
lsm5, line 21: shouldn't this be $2

lsm5 (commented):
LGTM mostly, except a couple of issues

setup
rhatdan: No it is $2 for the script but this is passed in as $1 in the function.

PR #15, opened 2023-04-12
Spec file should require hirte-agent and podman > 2.5 be available

rhatdan:


rhatdan:
@lsm5 @alexlarsson PTAL

lsm5:
Just FYI, RE: `podman >= 4.5` in the upstream spec.rpkg, 

- That would work if people if have a default Fedora podman and force them to get a newer podman from somewhere.
- It doesn't really do much for podman installed from the copr, cause if you look at the `epoch:version` on the podman builds on the coprs, they are of the form `101:0.0.FOO` . 
- Regardless, given the epoch is 101, this PR change should still work for installation. And podman on the `rhcontainerbot/qm` copr repo is actually v4.5 so we're good.

PR #17, opened 2023-04-13
Unmask all /proc and /sys, so that qm can run containers

rhatdan:


setup
```diff
@@ -22,7 +22,7 @@ install() {
     . /etc/os-release
     mkdir -Z -p ${rootfs}
     dnf -y install --releasever=${VERSION_ID} --installroot ${rootfs} selinux-policy-targeted podman systemd hirte-agent --nogpgcheck
-    dnf -y update --installroot ${rootfs}
+    dnf -y update --installroot ${rootfs} --nogpgcheck
```
lsm5, line 25: Might be worth leaving a FIXME comment here to get rid of --nogpgcheck.

lsm5 (approved):
LGTM mostly. Might be good to set a reminder for --nogpgcheck though.

setup
rhatdan: Yes I need to talk to Alex about this to see if he has any ideas or potentially some installation people.

PR #18, opened 2023-04-13
Install qm.container in /usr/share/containers/systemd

rhatdan:
/etc/containers/systemd/qm.container is for administrators to modify

Also run containers in the qm with level s0


lsm5 (approved):
LGTM

PR #22, opened 2023-04-21
Add /dev/kvm device to qm, so VMs can be launched

rhatdan:
In RHIVOS we want to support running VMs and specifically Android within the QM, which requires /dev/kvm to be present.


rhatdan:
@slp @alexlarsson @lsm5 @pypingou PTAL

PR #23, opened 2023-04-21
Automatically setup the hirte-agent

rhatdan:
Attempt to first copy the /etc/hirte/agent.conf from the host into the hirte environmnet, and add qm. to the front of the NodeName field.

If the NodeName field does not exists or is emtpy, then add qm.HOSTNAME as the NodeName.

If the NodeName is qm., then add HOSTNAME to the end.

### DIFF
--- a/setup
+++ b/setup
@@ -1,4 +1,4 @@
-#! /bin/sh -ex
+#! /bin/sh -e
 #
 # This setup script will install an OS environment by default into
 # /usr/share/qm/rootfs and create a Podman quadlet containerized environment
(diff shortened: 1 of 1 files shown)


rhatdan:
@alexlarsson @lsm5 @pypingou @mperina @engelmi PTAL

rhatdan:
@ygalblum PTAL

lsm5 (commented):
LGTM AFAICT, but I'll defer to a hirte expert.

ygalblum (approved):
LGTM

PR #26, opened 2023-04-27
quadlet: make `/dev/kvm` optional

lsm5:
This is required on environments like VMs where /dev/kvm is not present.

@rhatdan @alexlarsson PTAL

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -12,7 +12,7 @@ Slice=QM.slice
 
 [Container]
 AddCapability=all
-AddDevice=/dev/kvm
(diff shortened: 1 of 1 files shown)


lsm5:
support for this in quadlet is already merged https://github.com/containers/podman/pull/18378

lsm5:
@rhatdan I've verified that this change does work on c9s now.

lsm5:
> @rhatdan I've verified that this change does work on c9s now.

on c9s on gce with no `/dev/kvm`

PR #31, opened 2023-05-05
quadlet: minimal sane default for AllowedCPUs

lsm5:
Set `AllowedCPUs=` to a minimal sane default which should work for most if not all users.

RPM packaging can optimize this for the env in postinstall.

@rhatdan PTAL


rhatdan:
LGTM

PR #36, opened 2023-05-09
Packit: Add downstream Fedora packaging tasks

lsm5:
Get rid of `qm.spec.rpkg` in favour of
`rpm/qm.spec` which gets synced with fedora dist-git on every upstream release. The version in the new spec file is set to `0` by default and gets updated by packit automatically on every packit task.

For local manual rpm builds using the spec, the helper script in the `rpm/` subdir will update the Version field with the latest version found in the upstream repo.

Packit will automatically create a PR on fedora dist-git on every new upstream release. A sample PR will look like:
https://src.fedoraproject.org/rpms/container-selinux/pull-request/10#

A dry run for this can be triggered using:
`$ packit propose-downstream --local-content`

To run this command locally, you would need to have your packit user-configuration-file set.
Ref: https://packit.dev/docs/configuration/#user-configuration-file

along with a fedora api key created at:
https://src.fedoraproject.org/settings#nav-api-tab with sufficient ACLs.

### DIFF
--- a/.packit.sh
+++ b/.packit.sh
@@ -6,6 +6,9 @@
 
 set -eo pipefail
 
+# Set path to rpm spec file
+SPEC_FILE=rpm/qm.spec
+
 # Get Version from HEAD
 HEAD_VERSION=$(grep '^policy_module' qm.te | sed 's/[^0-9.]//g')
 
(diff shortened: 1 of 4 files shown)


lsm5:
@rhatdan PTAL, this should be good to go.

PR #41, opened 2023-05-13
e2e: add validations for calls 

dougsland:
The script was ready but missing to check if the commands really executed correctly. This patch adds such ability in all calls.


dougsland:
@rhatdan @lsm5 PTAL.

rhatdan:
LGTM

PR #47, opened 2023-05-16
Build selinux on platforms without user_namespace class

rhatdan:


lsm5:
@rhatdan could you please also remove the user_namespace sed line from `rpm/qm.spec` ? That'll verify if the change works well with the rpm.

lsm5:
> @rhatdan could you please also remove the user_namespace sed line from `rpm/qm.spec` ? That'll verify if the change works well with the rpm.

along with `no_user_namespace` conditionals at lines 15-21 in the spec.

dougsland:
removing the sed in the spec +1. Otherwise, the makefile solution is great solution.

dougsland:
Otherwise, we can merge now and later remove in a single commit the sed from spec @lsm5

dougsland:
Solves: https://github.com/containers/qm/pull/47

lsm5:
> Otherwise, we can merge now and later remove in a single commit the sed from spec @lsm5

either way works. But, I hope the spec file change verification lands before we cut a new release. Might be just a nit but better to have that change in when we have the chance.

lsm5 (approved):
LGTM once build passes and PR rebased on main.

lsm5:
copr builds seem to have failed because of merge conflicts. So, I think rebasing on main will take care of that too.

PR #48, opened 2023-05-16
RPM: adjust podman dependency epoch

lsm5:
Ref: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-9edad6c8f4


lsm5:
> /lgtm

don't think that works here yet. We would need to add the openshift mergebot here.

PR #50, opened 2023-05-16
qm.te: Bump version

lsm5:
@rhatdan @dougsland PTAL

@rhatdan would be nice to cut v0.2.1, but your call.


lsm5:
Gets the correct version now: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/5925262/

rhatdan:
LGTM

PR #51, opened 2023-05-16
Packit: check for version consistency in VERSION and qm.te

lsm5:
The version is mentioned in two places in the repo. This commit will check for version consistency before proceeding with packit tasks and fail on mismatch.


lsm5:
@dougsland @rhatdan PTAL

rhatdan:
LGTM

PR #52, opened 2023-05-16
github: add dougsland as reviewer

dougsland:


dougsland:
@lsm5 thanks, here the patch. Let's wait Dan's review too.

rhatdan (approved):
LGTM

PR #55, opened 2023-05-17
Quick doc fix in the setup script

pypingou:
The lines documenting what the script does (at the beginning of the script) were pointing to the wrong path/folder being created by this script.

This patch fixes that path to be the correct one

Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>


lsm5 (approved):
LGTM

PR #57, opened 2023-05-17
Update qm /etc/agent/agent.conf that when the qm.service is started

rhatdan:
This pull request will cause the /usr/lib/qm/rootfs/etc/hirte/agent.conf to be created on every qm.service start.  This way if the admin changes the /etc/hirte/agent.conf on the host, the equivalent inside of the qm will match with the exception that the nodename is prefixed with qm.

If the nodename in the hosts /etc/hirte/agent.conf is changed to foobar, then the qm's /etc/hirte/agent.conf will have the same settings except the node name will be qm.foobar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 The main purpose of this package is allow users to setup an environment which
 prevents applications and container tools from interfering with other all
-other processes on the system.
+other processes on the system. For example ASIL (Automotive Safety Integrity Level)
+environments.
 
 The QM environment uses containerization tools like cgoups, namespaces, and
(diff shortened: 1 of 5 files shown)


dougsland:
LGTM

lsm5 (approved):
LGTM

pypingou:
Worth a try, we should document that behavior somewhere though as it means we won't be able to tweak the `agent.conf` in the QM side any longer (we could tweak the conf by only via `agent.conf.d`)

rhatdan:
Yes or modify the qm.container.

PR #59, opened 2023-05-17
Packit: Disable downstream epel9 tasks

lsm5:
Podman v4.5 and hirte are not available yet in epel9 environments so
those targets would need to be disabled until then.

This doesn't affect the copr tasks as they don't depend on default
distro repos to fetch the latest packages.

Ref: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-77d64cf134#comment-3026157

@rhatdan @dougsland PTAL

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -67,18 +67,20 @@ jobs:
     dist_git_branches:
       - fedora-rawhide
       - fedora-38
-      - epel-9
+      # Disable epel-9 until hirte and podman 4.5 are available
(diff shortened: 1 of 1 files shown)


lsm5:
> I was able to download qm from epel-testing after installing podman on the stream's koji. Hirte got pulled in from the copr repo enabled on the vm and that seems to have worked. Having the qm build on epel-testing was quite helpful for this. Could we maybe build for epel-9 without pushing it?

doesn't epel9 have mass rebuilds? Mass rebuilds would mean trouble. Why not use the `rhcontainerbot/qm` copr instead for rhel9 envs ?

lsm5:
closing in favor of an upcoming PR

PR #61, opened 2023-05-17
Fix handling of /var and /etc within container

rhatdan:
With the current code using builtin volumes users on the host system can not modify the ROOTFS/var and ROOTFS/etc directories.
This PR just remounts these directories on top of themselves read/write allowing users from the host to modify the directories and the processes inside of the container to modify them.
Both the host and the QM will be able to see the modifications.

Added a ROOTFS=/usr/lib/qm/rootfs environment variable to make the quadlet easier to understand.

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -9,6 +9,7 @@ MemorySwapMax=0
 OOMScoreAdjust=500 
 Restart=always
 Slice=QM.slice
+Environment=ROOTFS=/usr/lib/qm/rootfs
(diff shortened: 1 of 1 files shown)


rhatdan:
@pypingou @alexlarsson @imcleod @lsm5 PTAL

rhatdan:
@dougsland PTAL

dougsland:
LGTM

PR #62, opened 2023-05-18
Packit: Temporarily disable bodhi for epel9

lsm5:
Podman v4.5 and hirte-agent are not yet available on epel9 so this commit removes those dependencies for epel9 and leaves it to the user to fetch those.

Bodhi updates are also disabled so the built package never gets submitted to any yum repo.

Refs:
1. https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-77d64cf134#comment-3026157
2. https://github.com/containers/qm/pull/59#issuecomment-1551691316

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -81,4 +81,6 @@ jobs:
     dist_git_branches:
       # rawhide updates are created automatically
       - fedora-38
-      - epel-9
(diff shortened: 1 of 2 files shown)


lsm5:
@pypingou @rhatdan @dougsland PTAL

PR #64, opened 2023-05-18
Bump to v0.4.0

rhatdan:
We never really released v0.3.0, but a lot of good features got added shortly after we started to release.

So I bumped the release again.


lsm5 (approved):
LGTM

PR #65, opened 2023-05-18
 RPM: account for nproc == 1 and update correct quadlet file

lsm5:
@rhatdan @dougsland @pypingou PTAL

### DIFF
--- a/rpm/qm.spec
+++ b/rpm/qm.spec
@@ -12,6 +12,14 @@
 # Format must contain '$x' somewhere to do anything useful
 %global _format() export %1=""; for x in %{modulenames}; do %1+=%2; %%1+=" "; done;
 
+# copr_username is only set on copr environments, not on others like koji
+# Check if copr is owned by rhcontainerbot
+%if "%{?copr_username}" != "rhcontainerbot"
+%bcond_with copr
+%else
+%bcond_without copr
+%endif
+
 %if 0%{?rhel}
 %bcond_with podman_45
 %bcond_with hirte_agent
@@ -21,6 +29,10 @@
 %endif
 
 Name: qm
+# Set different Epochs for copr and koji
+%if %{with copr}
+Epoch: 101
+%endif
 # Keep Version in upstream specfile at 0. It will be automatically set
 # to the correct value by Packit for copr and koji builds.
 # IGNORE this comment if you're looking at it in dist-git.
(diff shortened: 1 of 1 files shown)


dougsland (approved):
looks sane to me
/lgtm

rpm/qm.spec
```diff
@@ -83,7 +83,7 @@ sed -i 's/^install: man all/install:/' Makefile
 
 # Set AllowedCPUs in quadlet file
 NPROC=$(nproc)
-if [[ $NPROC == 2 ]]; then
+if [[ $NPROC <= 2 ]]; then
```
rhatdan, line 86: I am not sure if this will work correctly, IE Will 1 work when there is only 1 cpu?  Should it be 0?

rhatdan:
I also think this is wrong.  We should be doing
```
sed  "s/^AllowedCPUs=.*/AllowedCPUs=$ALLOWED_CPUS/" %{_datadir}/containers/systemd/qm.container > %{_sysconfdir}/containers/systemd/qm.container 
```
Currently after install I see.

 rpm -qV qm
.M.......  g /usr/lib/qm
.M.......  g /usr/lib/qm/rootfs
S.5....T.    /usr/share/containers/systemd/qm.container

rpm/qm.spec
lsm5: oh you're right, my bad. sorry

lsm5:
> I also think this is wrong. We should be doing
> 
> ```
> sed  "s/^AllowedCPUs=.*/AllowedCPUs=$ALLOWED_CPUS/" %{_datadir}/containers/systemd/qm.container > %{_sysconfdir}/containers/systemd/qm.container 
> ```
> 
> Currently after install I see.
> 
> rpm -qV qm .M....... g /usr/lib/qm .M....... g /usr/lib/qm/rootfs S.5....T. /usr/share/containers/systemd/qm.container

Remind me where the file is supposed to be? 

> I also think this is wrong. We should be doing
> 
> ```
> sed  "s/^AllowedCPUs=.*/AllowedCPUs=$ALLOWED_CPUS/" %{_datadir}/containers/systemd/qm.container > %{_sysconfdir}/containers/systemd/qm.container 
> ```
> 
> Currently after install I see.
> 
> rpm -qV qm .M....... g /usr/lib/qm .M....... g /usr/lib/qm/rootfs S.5....T. /usr/share/containers/systemd/qm.container

ack, I'll change that. Thanks

lsm5:
@rhatdan PTAL

lsm5:
need to fix the copr epoch issue as well.

lsm5:
with the latest change, copr builds have the Epoch: 101 and official koji builds continue with Epoch: 0.

rhatdan:
LGTM

rpm/qm.spec
```diff
@@ -21,6 +29,10 @@
 %endif
 
 Name: qm
+# Set different Epochs for copr and koji
+%if %{with copr}
+Epoch: 101
+%endif
```
pypingou, line 35: Hm, what's the goal of playing with the epoch here? there are fairly high costs associated with epochs :/

Issue #84, opened 2023-06-11
qm.te:3:ERROR 'unknown class user_namespace' at token ';' on line 4689:


dougsland:
CI is failing in this one:
````
Finish: build setup for qm-0.4.2-1.20230611184105946704.pr83.25.gc293cec.el9.src.rpm
Start: rpmbuild qm-0.4.2-1.20230611184105946704.pr83.25.gc293cec.el9.src.rpm
error: %changelog entries must start with *
Building target platforms: x86_64
Building for target x86_64
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.n3K76Z
+ umask 022
+ cd /builddir/build/BUILD
+ cd /builddir/build/BUILD
+ rm -rf qm-0.4.2
+ /usr/bin/gzip -dc /builddir/build/SOURCES/qm-0.4.2.tar.gz
+ /usr/bin/tar -xof -
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ cd qm-0.4.2
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
+ /usr/bin/git init -q
+ /usr/bin/git config user.name rpm-build
+ /usr/bin/git config user.email '<rpm-build>'
+ /usr/bin/git config gc.auto 0
+ /usr/bin/git add --force .
+ /usr/bin/git commit -q --allow-empty -a --author 'rpm-build <rpm-build>' -m 'qm-0.4.2 base'
+ sed -i 's/^install: man all/install:/' Makefile
+ RPM_EC=0
++ jobs -p
+ exit 0
Executing(%build): /bin/sh -e /var/tmp/rpm-tmp.lhwK3y
+ umask 022
+ cd /builddir/build/BUILD
+ cd qm-0.4.2
+ /usr/bin/make all
mkdir -p tmp; cp qm.* tmp/
sed -i /"user_namespace exists"/d tmp/qm.if
make -C tmp -f /usr/share/selinux/devel/Makefile qm.pp
make[1]: Entering directory '/builddir/build/BUILD/qm-0.4.2/tmp'
Compiling targeted qm module
qm.te:3:ERROR 'unknown class user_namespace' at token ';' on line 4689:
	allow qm_t self:user_namespace all_user_namespace_perms;
#line 3
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
make[1]: *** [/usr/share/selinux/devel/include/Makefile:157: tmp/qm.mod] Error 1
[...]

PR #90, opened 2023-06-14
makefile: remove str exists from sed

dougsland:
Resolves: https://github.com/containers/qm/issues/84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ DATADIR ?= $(PREFIX)/share
 LIBDIR ?= $(PREFIX)/lib
 SYSCONFDIR?=/etc
 QMDIR=/usr/lib/qm
-USER_NAMESPACE="user_namespace exists"
+USER_NAMESPACE="user_namespace"
 ifeq ($(wildcard /sys/fs/selinux/class/user_namespace),)
 	USER_NAMESPACE="user_namespace"
 endif


dougsland:
@rhatdan I believe this is the fix, let's see the CI result.

Makefile
```diff
@@ -6,7 +6,7 @@ DATADIR ?= $(PREFIX)/share
 LIBDIR ?= $(PREFIX)/lib
 SYSCONFDIR?=/etc
 QMDIR=/usr/lib/qm
-USER_NAMESPACE="user_namespace exists"
```
rhatdan, line 9: This is wrong, we want to USER_NAMESPACE to not be removed which means the sed will not remove it.

rhatdan: This change would cause it to always be removed.

Yarboa (approved):
LGTM

Makefile
dougsland: @rhatdan hum, that's why the spec always worked. The spec had this sed. Also, I cannot see `user_namespace exists` for Centos9 or Fedora (in the generated qm.if). Any other thing to me to dig?

dougsland: @lsm5  I don't see user_namespace exists :(

# cat /etc/redhat-release
CentOS Stream release 9

# make qm.pp

# grep -rni user_namespace qm.if
135:	allow $1_t self:user_namespace all_user_namespace_perms;
437:	allow $1_container_domain self:user_namespace create;

# rpm -qa | grep -i selinux
libselinux-3.5-1.el9.x86_64
python3-libselinux-3.5-1.el9.x86_64
libselinux-utils-3.5-1.el9.x86_64
selinux-policy-38.1.14-1.el9.noarch
selinux-policy-targeted-38.1.14-1.el9.noarch
container-selinux-2.218.0-1.el9.noarch
rpm-plugin-selinux-4.16.1.3-23.el9.x86_64
selinux-policy-devel-38.1.14-1.el9.noarch

rhatdan: Right, which means the sed statement will remove nothing.  The problem is sed will not work with "" So we needed to state something.  The idea is to only sed user_namespace if the class Does not exists

rhatdan:
The question is what didn't this remove user_namespace on centos systems?

PR #91, opened 2023-06-15
TEST

dougsland:


lsm5:
@dougsland i'm checking with the copr folks about the c9s environment..

lsm5:
@dougsland @rhatdan so looks like the copr build tasks for c9s happen in a mock chroot on an actual f38 builder. So, checking for `/sys/fs/selinux/class/user_namespace` may be sufficient in Makefile for local builds, but we'll also need to remove user_namespace explicitly in `rpm/qm.spec` to make CI happy.

lsm5:
PTAL: https://github.com/containers/qm/pull/93

PR #93, opened 2023-06-15
test

lsm5:


lsm5:
@dougsland builds passed here. PTAL

lsm5:
we could either do it this way or perhaps check in the Makefile for `/etc/redhat-release` as well

lsm5:
@Yarboa @rhatdan PTAL

lsm5:
@dougsland @Yarboa feel free to merge this to get things unblocked. Only affects rpms on c9s (if at all). @rhatdan might be occupied with devconf.cz .

dougsland:
@lsm5 could you please confirm if this helps keep the logic in the makefile:
```
 @if grep -q "CentOS Stream 9" /etc/os-release; then \
                echo "Host is CentOS Stream 9"; \
        else \
                echo "Host is not CentOS Stream 9"; \
        fi
```

lsm5:
> > @lsm5 could you please confirm if this helps keep the logic in the makefile:
> > ```
> >  @if grep -q "CentOS Stream 9" /etc/os-release; then \
> >                 echo "Host is CentOS Stream 9"; \
> >         else \
> >                 echo "Host is not CentOS Stream 9"; \
> >         fi
> > ```
> 
> Of course, using TAB as space. It's bad indentation here.

RE: Makefile, do we see this condition hitting environments other than CI / copr / koji environments? I got no strong opinion on this btw.

grepping for that string works, but may want to grep for `Red Hat Enterprise Linux 9` if you want it to be more generic.

dougsland:
> > > @lsm5 could you please confirm if this helps keep the logic in the makefile:
> > > ```
> > >  @if grep -q "CentOS Stream 9" /etc/os-release; then \
> > >                 echo "Host is CentOS Stream 9"; \
> > >         else \
> > >                 echo "Host is not CentOS Stream 9"; \
> > >         fi
> > > ```
> > 
> > 
> > Of course, using TAB as space. It's bad indentation here.
> 
> RE: Makefile, do we see this condition hitting environments other than CI / copr / koji environments? I got no strong opinion on this btw.
> 
> grepping for that string works, but may want to grep for `Red Hat Enterprise Linux 9` if you want it to be more generic.

sure

dougsland:
@lsm5 @Yarboa @rhatdan finally made via Makefile too: https://github.com/containers/qm/pull/94
I am okay to merge both, have the spec don't hurt but should not be required.

rhatdan:
LGTM

PR #94, opened 2023-06-21
make: add validation for rhel9 or lower

dougsland:


rhatdan:
@lsm5 Can't we make this cleaner with RPM macros?

rhatdan:
Ok so this is not in a rpm spec.  I still want to know why the original check is not working?

Can we just check for the existence of `/sys/fs/selinux/class/user_namespace`

lsm5:
> Ok so this is not in a rpm spec. I still want to know why the original check is not working?
> 
> Can we just check for the existence of `/sys/fs/selinux/class/user_namespace`

@dougsland if the spec file patch was merged, do we still need this? Would people building directly from Makefile be affected without this patch?

dougsland:
> > Ok so this is not in a rpm spec. I still want to know why the original check is not working?
> > Can we just check for the existence of `/sys/fs/selinux/class/user_namespace`
> 
> @dougsland if the spec file patch was merged, do we still need this? Would people building directly from
> Makefile be affected without this patch?

I think so. Even that we are a rpm shop only, it's always good to make the project friendly to others distros that would like to use it.

dougsland:
> Ok so this is not in a rpm spec. I still want to know why the original check is not working?
> 
> Can we just check for the existence of `/sys/fs/selinux/class/user_namespace`

In the Github Actions Virtual Machine it exists in my local podman CentOS Stream 9 nop.

lsm5:
@dougsland ack, sgtm. But to make sure this works, can you also get rid of all the conditionals in rpm/qm.spec to handle rhel 9 ? The centos 9 copr builds should pass with those removed.

rhatdan:
LGTM
@lsm5 PTAL

lsm5:
@dougsland squash commits and we're good to merge.

PR #95, opened 2023-06-21
e2e: add gh tool

dougsland:
GitHub tool helps to bring patch easy for the test scenarios.

### DIFF
--- a/tests/e2e/lib/ContainerFile.template
+++ b/tests/e2e/lib/ContainerFile.template
@@ -6,10 +6,12 @@ RUN dnf install -y \
 
 RUN dnf config-manager -y --set-enabled crb
 RUN dnf -y copr enable rhcontainerbot/qm centos-stream-9
+RUN dnf config-manager --add-repo https://cli.github.com/packages/rpm/gh-cli.repo
 
 RUN dnf install -y \
 	clang-tools-extra \
         codespell \
+	gh \
         bzip2 \
         gcc \
         meson \


dougsland:
the error not related to the patch. See: https://github.com/containers/qm/pull/94

rhatdan:
LGTM

lsm5:
> GitHub tool helps to bring patch easy for the test scenarios.

sorry, could you please explain this further? Maybe add a test case for this? Not opposed to this, but just curious why it's needed.

dougsland:
> > GitHub tool helps to bring patch easy for the test scenarios.
> 
> sorry, could you please explain this further? Maybe add a test case for this? Not opposed to this, but just
> curious why it's needed.

Sure, I use the e2e tool locally as well. I have being testing hirte with patches from different developers and it's pretty quick to copy the command line from the `< > Code` button in Github to bring the PR to the environment and test. Example: `gh pr checkout PR_NUMBER`.  If this is bad, feel free to drop the PR. Just trying to make an auto environment for development and testing via qm e2e tool.

lsm5:
> > > GitHub tool helps to bring patch easy for the test scenarios.
> > 
> > 
> > sorry, could you please explain this further? Maybe add a test case for this? Not opposed to this, but just
> > curious why it's needed.
> 
> Sure, I use the e2e tool locally as well. I have being testing hirte with patches from different developers and it's pretty quick to copy the command line from the `< > Code` button in Github to bring the PR to the environment and test. Example: `gh pr checkout PR_NUMBER`. If this is bad, feel free to drop the PR. Just trying to make an auto environment for development and testing via qm e2e tool.

Yes, I use `gh pr ..` too, but maybe we should let the user handle this themselves than pull in a dependency? If it's not really needed by CI itself, then maybe we can have this as an addon somewhere, maybe a couple of sed lines to update the Containerfile.template via a script in `contrib/` directory the human could run.

dougsland:
> > > > GitHub tool helps to bring patch easy for the test scenarios.
> > > 
> > > 
> > > sorry, could you please explain this further? Maybe add a test case for this? Not opposed to this, but just
> > > curious why it's needed.
> > 
> > 
> > Sure, I use the e2e tool locally as well. I have being testing hirte with patches from different developers and it's pretty quick to copy the command line from the `< > Code` button in Github to bring the PR to the environment and test. Example: `gh pr checkout PR_NUMBER`. If this is bad, feel free to drop the PR. Just trying to make an auto environment for development and testing via qm e2e tool.
> 
> Yes, I use `gh pr ..` too, but maybe we should let the user handle this themselves than pull in a dependency?
> If it's not really needed by CI itself, then maybe we can have this as an addon somewhere, maybe a couple of
> sed lines to update the Containerfile.template via a script in `contrib/` directory the human could run.

I don't think it will change anything in CI but OK, completely understand.  I will handle that myself.

PR #96, opened 2023-06-22
RPM: ensure user_namespace removed only on rhel9

lsm5:


lsm5:
@rhatdan @dougsland PTAL.

lsm5:
This may not be needed if things are handled in Makefile: https://github.com/containers/qm/pull/94

rhatdan:
LGTM

PR #100, opened 2023-07-07
Make %ghost own the file created in %post

pypingou:
Otherwise we end up with a file in /etc/containers/systemd that isn't linked in the RPM database to the qm package making people wonder where this is coming from.


lsm5 (approved):
LGTM

@rhatdan PTAL

rhatdan:
Oops
LGTM

PR #111, opened 2023-07-17
fix connectivity test shellcheck

sandrobonazzola:
fixed issues detected by shellcheck on qm connectivity test.


dougsland:
LGTM

PR #115, opened 2023-07-19
start qm.service at boot

### LINKED ISSUE #105: qm service not started after reboot
On a clean CentOS Stream 9, after completing qm setup everything is up and running.

But after rebooting the host, the `qm.service` doesn't start automatically: in order to get it up `systemctl start qm` is needed.

sandrobonazzola:
Added Install section as per
https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html in order to get the container up after reboot.

Fixes #105

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -1,3 +1,6 @@
+[Install]
+WantedBy=default.target
+
 [Service]
 AllowedCPUs=1
 CPUWeight=50


dougsland (approved):
LGTM

PR #116, opened 2023-07-19
stop qm.service only on real install

### LINKED ISSUE #106: setup kills qm.service startup via ExecStartPre
After #102 the qm service can't start anymore due to:

```
× qm.service
     Loaded: loaded (/etc/containers/systemd/qm.container; generated)
     Active: failed (Result: signal) since Mon 2023-07-17 09:34:59 CEST; 5s ago
    Process: 1588 ExecStartPre=/usr/share/qm/setup hirte-agent (code=killed, signal=TERM)
    Process: 1590 ExecStopPost=/usr/bin/podman rm -f -i --cidfile=/run/qm.cid (code=exited, status=0/SUCCESS)
        CPU: 33ms
```

 `/usr/share/qm/setup hirte-agent` gets killed probably because setup calls `systemctl stop qm.service 2>/dev/null || true`
just before `hirteSetup` gets executed.

please note #105 is not related to this issue, commenting out the ExecStartPre config still cause the service to not be started at reboot.

sandrobonazzola:
previously qm.service was stopped always, including when reconfiguring hirte-agent at qm.service start.

Now qm.service is stopped only if a real install is being executed.

Fixes #106

### DIFF
--- a/setup
+++ b/setup
@@ -83,7 +83,9 @@ case "$1" in
 	hirteSetup "${ROOTFS}"
 	;;
     *)
-	systemctl stop qm.service 2>/dev/null || true
+	if systemctl is-active --quiet "qm.service" ; then
+		systemctl stop qm.service
+	fi
 	install "${ROOTFS}"
 	replaceIDs /etc/subuid qmcontainers ${qmContainerIDs}
 	replaceIDs /etc/subgid qmcontainers ${qmContainerIDs}
--- a/setup
+++ b/setup
@@ -23,8 +23,6 @@ INSTALLDIR="$1"
 ROOTFS="$2"
 [ -n "${ROOTFS}" ] || ROOTFS=/usr/lib/qm/rootfs
 
-systemctl stop qm.service 2>/dev/null || true
-
 hirteSetup() {
     ROOTFS=$1
     if test ! -f "${ROOTFS}${AGENTCONF}"; then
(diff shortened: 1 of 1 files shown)


setup
```diff
@@ -85,6 +83,7 @@ case "$1" in
 	hirteSetup "${ROOTFS}"
 	;;
     *)
+	systemctl stop qm.service 2>/dev/null || true
```
dougsland, line 86: Just a nit comment, as changing the line, probably good to add extra lines to verify the service is running and stop it.
 
```
if systemctl is-active --quiet "qm.service" ; then
  systemctl stop qm.service 2>/dev/null || true
fi
```

sandrobonazzola: pushed a commit addressing this comment. Dropped the trailing ` 2>/dev/null || true` as now we are testing if the service is active and if it fails to stop it's nice to know why.

setup
```diff
@@ -83,7 +83,9 @@ case "$1" in
 	hirteSetup "${ROOTFS}"
 	;;
     *)
-	systemctl stop qm.service 2>/dev/null || true
+	if systemctl is-active --quiet "qm.service" ; then
+		systemctl stop qm.service 2>/dev/null
```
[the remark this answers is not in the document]
sandrobonazzola: qm.service doesn't output on stop; I just forgot to drop the stderr redirect :-) pushing again

dougsland:
the CI/CD error not related to the change, we will need to investigate.
```
rror: no container with name or ID "qm" found: no such container[0m
[33m[15:36:49] [W] [worker_0] [stderr] [CentOS-Stream-9:x86_64:/tests/e2e/tier-0]         Error: no container with name or ID "qm" found: no such container[0m
[33m[15:36:49] [W] [worker_0] [stderr] [CentOS-Stream-9:x86_64:/tests/e2e/tier-0]         Error: no container with name or ID "qm" found: no such container[0m
[33m[15:36:49] [W] [worker_0] [stderr] [CentOS-Stream-9:x86_64:/tests/e2e/tier-0]         Error: no container with name or ID "qm" found: no such container[0m
```

PR #124, opened 2023-07-21
RPM: autospec macros not defined yet in EL9

lsm5:


dougsland:
beautiful solution, waiting CI pass.
/cc @sandrobonazzola

dougsland:
@sandrobonazzola can we close an issue now?

pypingou:
The way I've done this was to simply import the src.rpm from epel9 when building for el9, that allows to avoid this conditional

PR #131, opened 2023-07-26
Fix hirtectl config

Yarboa:
During inspection of systemctl start qm failures
It seems that config files were not set properly due to regexp inaccuracies


dougsland (approved):
Thanks Yaniv, it's related to the new `hirte config load from sources` feature. Great catch!

PR #134, opened 2023-07-26
test

dougsland:


rhatdan:
I have no idea what this does other then add a comment.
LGTM

dougsland:
@rhatdan sorry, I am testing CI/CD with the last changes but looks like we need to wait a new rpm be published in the copr so CI/CD pull it and do the tests. Tomorrow I expect CI/CD green, I will also check with @lsm5 how to trigger a fresh build in case like that.

Issue #103, opened 2023-07-14
invalid request descriptor after setup completed


sandrobonazzola:
On a clean CentOS Stream 9 minimal VM I setup qm.

I can see it up:
```console
# podman exec -ti qm ps -eZ
LABEL                               PID TTY          TIME CMD
system_u:system_r:qm_t:s0             1 ?        00:00:00 systemd
system_u:system_r:qm_t:s0            24 ?        00:00:00 systemd-journal
system_u:system_r:qm_t:s0            33 ?        00:00:00 hirte-agent
system_u:system_r:qm_t:s0            34 ?        00:00:00 systemd-logind
system_u:system_r:qm_t:s0            36 ?        00:00:00 dbus-broker-lau
system_u:system_r:qm_t:s0            37 ?        00:00:00 dbus-broker
system_u:system_r:qm_t:s0           161 pts/0    00:00:00 ps
```

I followed the demo steps and imported the quadlet for running httpd container within qm.

My host has hostname `centos9.home` and the setup created :
```console
# cat /usr/lib/qm/rootfs/etc/hirte/agent.conf
[hirte-agent]
NodeName=qm.centos9.home
```

On the host I have:
```console
# cat /etc/hosts 
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6

# hostname
centos9.home
```

When I try to run the quadlet I have:
```console
# hirtectl restart qm.centos9.home myquadlet.service
Failed to issue method call: Unknown object '/org/containers/hirte/node/qm_2ecentos9_2ehome'.
Call to manager failed: Invalid request descriptor
```

I think some configuration is missing on the host as I don't see any configuration for telling the manager where to look for `qm.centos9.home`:
```console
# ls -R /etc/hirte/
/etc/hirte/:
agent.conf.d  hirte.conf.d

/etc/hirte/agent.conf.d:
README.md

/etc/hirte/hirte.conf.d:
README.md
```

sandrobonazzola:
Something like:
```console
# cat /etc/hirte/hirte.conf.d/00-allowed-nodes.conf 
[hirte]
AllowedNodeNames=qm.centos9.home
```
worked for me

dougsland:
> Something like:
> 
> ```
> # cat /etc/hirte/hirte.conf.d/00-allowed-nodes.conf 
> [hirte]
> AllowedNodeNames=qm.centos9.home
> ```
> 
[...]

PR #137, opened 2023-07-28
demo: general improvements

dougsland:
- check errors in commands
    - auto add node names into hirte config
    - general improvements in the demo script

    Resolves: https://github.com/containers/qm/issues/103


dougsland:
In a different patch, I want to create a README.md describing which environment is required. It will make life easier to users. A recent fedora should be always enough.

rhatdan:
SGTM

PR #147, opened 2023-08-05
e2e: remove commented line in the AllowedNodeNames

dougsland:
In this moment of the deploy, this option is not commented. Fixing the sed.


dougsland:
all checks green, making it fy.

PR #149, opened 2023-08-05
e2e: run-test-e2e add --number-of-nodes

dougsland:
To test hirte and qm in scale we will need a way to easily create 'N' nodes.

Example for creating 8 nodes with 8 qm containers inside:
   ./run-test-e2e --number-of-node=8


dougsland:
all green, merging.

PR #150, opened 2023-08-05
CODEOWNERS: Yariv and Sandro

dougsland:
- Yariv has being helping us in the QE side
- Sandro has being helping us in the tests for the virtualization platform
   and fedora packaging

Both understand and uses QM in different ways, if possible, i would like to add them as owners and reviewers of the code to keep our quality high.


Yarboa (approved):
LGTM

PR #151, opened 2023-08-05
e2e: update lib/container to add node'N' nodes

dougsland:
Now that we have a ability to create 'N' nodes
we need to add it to AllowedNodeNames too.


dougsland:
all green, merging.

PR #160, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not
properly isolate PRs from each other.

To avoid that, change the copr_build configuration to use the packit
default COPRs, which are specific to the particular PR, and disappear
after a few weeks. Depending projects should only run against what
landed in qm/main i.e. the podman-next COPR.

Signed-off-by: Martin Pitt <mpitt@redhat.com>

----

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,43 +2,36 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-# Build targets can be found at:
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-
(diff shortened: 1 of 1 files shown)


rhatdan:
@lsm5 PTAL

lsm5 (approved):
LGTM

PR #166, opened 2023-08-12
e2e: add packages for tmt development

dougsland:

### DIFF
--- a/tests/e2e/lib/ContainerFile.template
+++ b/tests/e2e/lib/ContainerFile.template
@@ -15,6 +15,9 @@ RUN dnf install -y \
         meson \
 	hostname \
 	podman \
+	python3-podman \
+	python3-pytest \
+	python3-pytest-timeout \
 	sudo \
         iptables-services \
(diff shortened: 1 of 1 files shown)


dougsland:
all green.

lsm5 (commented):
Too late to review but anyway. I guess these will end up getting installed on all envs that depend on Containerfile.template ? May not be worth pulling extra deps unless absolutely needed.

dougsland:
> Too late to review but anyway. I guess these will end up getting installed on all envs that depend on
> Containerfile.template ? May not be worth pulling extra deps unless absolutely needed.

Never too late, there is `git revert` ;-). However, I knew this could bring people complaining, but I believe we should start thinking in a different way for these scripts. This became a test and development platform (and E2E) for hirte and qm. Adding few extra piles/Mbs of software to speedup things won't hurt or change our tests results.

A simple example of usage outside of E2E in github actions:
https://github.com/containers/bluechi/issues/434

Please let me know your thoughts.

PR #170, opened 2023-08-17
Packit: Enable fedora-39 targets and bodhi for epel-9

lsm5:
PTAL.

podman v4.6 and hirte are now available on C9S, so we should be ok. Will the hirte / bluechi name be an issue?

FYI, the distro targets will become a lot easier to manage once fedora-37 is EOL.


lsm5:
@rhatdan @dougsland @sandrobonazzola @Yarboa PTAL.

rhatdan:
LGTM

PR #173, opened 2023-08-19
e2e: add pre-commit tool to validate code

dougsland:
pre-commit has several validations that can be used during development.


dougsland:
failing because of hirte still in the code, should be renamed to bluechi.

PR #186, opened 2023-08-29
Hostname

odra:
* adds `--hostname` to use a custom hostname
* enables the qm agent by creating the symlink directly if using `--skip-systemctl`.


dougsland (approved):
LGTM

PR #189, opened 2023-08-30
Update .packit.yaml

dougsland:
should not be needed to require rpkg


lsm5:
lgtm if builds are successful.

PR #197, opened 2023-09-03
e2e: add rust support

dougsland:


dougsland:
See also:
rust: Add BlueChi Rust module
https://github.com/containers/bluechi/pull/533

PR #198, opened 2023-09-05
e2e: add markdown-cli linter

dougsland:

### DIFF
--- a/tests/e2e/lib/ContainerFile.template
+++ b/tests/e2e/lib/ContainerFile.template
@@ -49,6 +49,7 @@ RUN dnf install -y \
 	python3-gobject \
 	python3-pip \
 	meson \
+	npm \
 	rpm-build \
 	ruby \
 	sed \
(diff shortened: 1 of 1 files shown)


sandrobonazzola (commented):
Is there any reason for running linting on the code during end to end tests?
I assume this should be covered by pre-commit already

dougsland:
> Is there any reason for running linting on the code during end to end tests?
> I assume this should be covered by pre-commit already

@sandrobonazzola 
bluechi project uses this markdown lint. Would be nice to have the same one they run there too.

https://github.com/containers/bluechi/blob/main/.github/workflows/lint.yml#L101

sandrobonazzola:
I won't be the one blocking this PR but IMHO if this container is going to be used for development it should live in another place, not in e2e. Having a bigger container for e2e testing means we consume more bandwidth, storage and time to run e2e testing. This consumes power and money, we should minimize the e2e test environment, not making it bigger.

dougsland:
> I won't be the one blocking this PR but IMHO if this container is going to be used for
> development it should live in another place, not in e2e. Having a bigger container for e2e
> testing means we consume more bandwidth, storage and time to run e2e testing. This 
>consumes power and money, we should minimize the e2e test environment, not making it
> bigger.

Thanks Sandro! We can always improve the solution. Let's merge now, I am aware that @Yarboa is working in some improvements regarding layers/images etc. Also, fell free to create an issue in the repo with the details and improvements you visualize so we can address it.

PR #199, opened 2023-09-05
e2e: Add dbus-devel package

dougsland:
dbus-devel provides libraries and header files needed for developing software that uses D-BUS.


dougsland:
See also: https://github.com/containers/bluechi/pull/544

PR #206, opened 2023-09-12
[WIP] limit the number of processes inside of the QM

rhatdan:
The default limit on number of processes within the container is 1024. This is far too low for QM. Selecting 10000 seems like a better number but really this needs to be set by the final distributor of the OS.

Requires https://github.com/containers/podman/pull/19955 to be in Podman.

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -21,6 +21,7 @@ AddDevice=-/dev/fuse
 ContainerName=qm
 Exec=/sbin/init
 Network=host
+PidsLimit=10000
 PodmanArgs=--security-opt label=nested --security-opt unmask=all
 ReadOnly=true
 Rootfs=${ROOTFS}


Yarboa (approved):
LGTM

rhatdan:
This needs to be revisited with podman 4.7 is released.

dougsland:
> This needs to be revisited with podman 4.7 is released.

@rhatdan should we also add a requirements in the rpm ? 
/cc @lsm5

dougsland:
LGTM anyway

dougsland:
btw, I faced similar issue already in the E2E when stressing the system with increase of the number of nodes:

```
    --number-of-nodes)
      NUMBER_OF_NODES="${2}"

      # avoid "too many files opened"
      sysctl fs.inotify.max_user_instances=$((${2}*100)) 1> /dev/null
```

The code is here:
https://github.com/containers/qm/blob/1e7bf2a3950c86b2cc338eb287c8f6d4656ef7fe/tests/e2e/run-test-e2e#L146-L147

rhatdan:
Changed to use Ssytemd mechanism, so closing this for now.

PR #207, opened 2023-09-12
limit the number of processes/tasks inside of the QM

rhatdan:
The default limit on number of processes within the container is 2048. This is far too low for QM. Selecting 50% of max seems like a better number but really this needs to be set by the final distributor of the OS.

Also set the maximum number of open file descriptors.


dougsland:
@rhatdan we also need to wait a specific version of podman for this one?

rhatdan:
No this one can go in now.

PR #208, opened 2023-09-12
[WIP] Add support to FFI

dougsland:
Just an initial code for FFI:

- Add C sources to eat memory (ASIL: 20%) (QM: 90%)
- ASIL container should create it's own file from template

### DIFF
--- a/tests/e2e/lib/ContainerFile.template
+++ b/tests/e2e/lib/ContainerFile.template
@@ -110,6 +110,10 @@ RUN sed -e '/^#ManagerPort=/s/^#//g' \
 	-e '/#LogTarget/s/^#//' \
 	-e "s/^#AllowedNodeNames=/AllowedNodeNames=control,node1/" \
 	-i /etc/bluechi/bluechi.conf
+
+RUN mkdir -p /root/tests/FFI
+COPY tools/FFI/memory/ASIL/20_percent_memory_eat.c /root/tests/FFI
+RUN cd /root/tests/FFI && gcc -o 20_percent_memory_eat 20_percent_memory_eat.c
 # [end] CONTROL - settings
 
 RUN echo 'Foobar!' | passwd --stdin root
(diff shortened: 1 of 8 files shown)


dougsland:
For the disks spaces we can even use fallocate tool, example:

```
fallocate -l 300G $filename
```

Yarboa (commented):
I would advise adding tests inside tmt tests and not inside the scripts
Consider copying the files while preparing the test,
under prepare of test.fmf 
You can also make the files in that stage

tests/e2e/lib/ContainerFile.template
```diff
@@ -110,6 +110,10 @@ RUN sed -e '/^#ManagerPort=/s/^#//g' \
 	-e '/#LogTarget/s/^#//' \
 	-e "s/^#AllowedNodeNames=/AllowedNodeNames=control,node1/" \
 	-i /etc/bluechi/bluechi.conf
+
+RUN mkdir -p /root/tests/FFI
+COPY tools/FFI/memory/ASIL/20_percent_memory_eat.c /root/tests/FFI
+RUN cd /root/tests/FFI && gcc -o 20_percent_memory_eat 20_percent_memory_eat.c
```
Yarboa, line 116: This part   should run under test prepare stage

tests/e2e/lib/container
```diff
@@ -112,6 +112,10 @@ create_qm_node() {
                 ' /etc/bluechi/agent.conf' >> ContainerFile.node"${nodeID}"
         echo "RUN systemctl enable bluechi-agent &> /dev/null" >> ContainerFile.node"${nodeID}"
 
+        echo "RUN mkdir -p /usr/lib/qm/rootfs/root/tests/FFI" >> ContainerFile.node"${nodeID}"
```
Yarboa, line 115: dame as above

dougsland:
based on the talk with @Yarboa , he will adapt this PR to something more aligned to QE. Thanks @Yarboa

PR #211, opened 2023-09-18
 Adding Dennis as default reviewer

Yarboa:


Yarboa:
@dougsland there is an error that users does'nt have write permission to qm repo,
Do they need to be members of containers project?

dougsland:
> @dougsland there is an error that users does'nt have write permission to qm repo, Do they need to be members of containers project?

yes, I can fix that.

dougsland:
> > @dougsland there is an error that users does'nt have write permission to qm repo, Do they need to be members of containers project?
> 
> yes, I can fix that.

sent invite to both, they need to accept.

dougsland:
LGTM

@lsm5 @rhatdan PTAL

rhatdan:
LGTM

rhatdan:
Welcome  @dennisbrendel @priyverm

PR #214, opened 2023-09-20
FFI: add exceed-file-descriptor tool

dougsland:

### DIFF
--- a/tests/e2e/tools/FFI/file-descriptors/QM/exceed-file-descriptor
+++ b/tests/e2e/tools/FFI/file-descriptors/QM/exceed-file-descriptor
@@ -1,0 +1,30 @@
+#!/usr/bin/python3
+
+import resource
+import os
+
+# Get the soft and hard limits for file descriptors
+soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+
+fd_limit = soft_limit
+
+description = "Description: Exceed the number " \
+    "of file descriptions (fd) in the QM system"
+
+expect_result = "Result: Should not affect ASIL" \
+    " environment when achieving \"Too many files opened\""
+
+print(f"\n{description}")
+print(f"{expect_result}\n")
+
+print(f"- Detected the number of fd in the system: {soft_limit} (soft limit)")
+print("- Trying to open as many file descriptors as possible...")
+
+# Attempt to open files until we reach/exceed the limit
+file_descriptors = []
+
+while len(file_descriptors) < fd_limit:
+    fd = os.open('/dev/null', os.O_RDONLY)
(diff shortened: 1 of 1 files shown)


dougsland:
@Yarboa @dennisbrendel @priyverm PTAL

dougsland:
```
     cp: cannot stat '/usr/share/bluechi/config/bluechi.conf': No such file or directory
        Error: building at STEP "RUN cp /usr/share/bluechi/config/bluechi.conf /etc/bluechi/": while running runtime: exit status 1
        Error: repository name must have at least one component
```

CI error not related to the change, it's related to the new config file bluechi-controller instead of bluechi.
Introduced by:  https://github.com/containers/bluechi/pull/570 We should fix CI soon.

dougsland:
> Not sure this one is worth it. ulimit nofile is not for the entire container only for a single process, As soon as the process forks/execs it gets another set of open file descriptors. I also found that there does not seem to be a system wide limit on fds, and if it was approached, I think the memory cgroups would fire off.
> 
> @alexlarsson @rhvgoyal WDYT?

It was listed as one of items to be addressed. 
```
Run out of Open FDs
Open FDs until you run out of FDs, Attempt to allocate a FD outside of the VM.
```

I don't mind cut this one off.

rhatdan:
@rhvgoyal @giuseppe is this something we should worry about?

dennisbrendel:
> I also found that there does not seem to be a system wide limit on fds, and if it was approached, I think the memory cgroups would fire off.

Memory consumption aside, could this have different negative effects on running applications? The kernel keeps some list with the fds, right? If that grows heavily, would it slow down other stuff?

giuseppe:
The memory cgroup limit should affect the file descriptor table too.  I've done a simple test with a program that opens a fd in a loop and the memory usage as tracked by the cgroup keeps increasing, from a minimum of 217088 bytes to a maximum of 430080 bytes when it reaches 1024 open file descriptors.

If you limit the number of processes in a container, as you should to prevent a PIDs exhaustion, then the RLIMIT_NOFILE  is affected by that, since you cannot have more than #PIDs * RLIMIT_NOFILE open file descriptors in the container

alexlarsson:
I don't see the point with this. File descriptor limits is purely an in-process 
protection. One process having many file descriptors does not really cause problems for other processes, only itself. Other than the memory use for the table that is, but as @giuseppe said, memory is limited in other ways.

dennisbrendel:
Thanks for the information!

I think it makes sense to have a verification in place that checks that the memory cgroup constrains such behavior (DOS prevention).

dougsland:
Thanks for the feedback, closing this one. @dennisbrendel maybe @giuseppe have such verification script done?

PR #215, opened 2023-09-21
Fix due to renaming of bluechid to controller

Yarboa:
Bluechid configuration files renamed from
bluechi.cong -> controller.conf
Updates scripts and Container files based on that


rhatdan:
LGTM

PR #217, opened 2023-09-26
Cleanup scripts

dennisbrendel:
Mostly cosmetic changes:
- be consistent in terms of tabs vs spaces
- consistent indentation depths
- use `local` variables
- only color output when the terminal signals support
- use variables for the color control sequences
- use heredoc instead of many `echo`s
- small fixes to the help message

### DIFF
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -17,8 +17,9 @@
 # along with this program; If not, see <http://www.gnu.org/licenses/>.
 
 get_ip_container() {
-    container_name="${1}"
-    container_interface="${2}"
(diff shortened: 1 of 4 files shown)


Yarboa (commented):
@dennisbrendel 
I see spaces changes, which shell checker do you use? 
Maybe we can add it to precommit hooks? 
https://github.com/containers/qm/blob/main/.pre-commit-config.yaml#L5

dennisbrendel:
I did not use a checker @Yarboa , just looked at the script with `vim` and saw the tab characters. Having a pre-commit hook in place makes sense.

Issue #223, opened 2023-09-28
setup: cp: cannot create regular file '/usr/lib/qm/rootfs/etc/containers/': Not a directory


dougsland:
Running setup script in my env (Fedora 38) had this failure:

```
# ./setup
setup has started...
  * rootfs: /usr/lib/qm/rootfs
  * rwetcfs: /etc/qm
  * rwvarfs: /var/qm
  * install dir: /usr/share/qm

Last metadata expiration check: 0:26:19 ago on Wed 27 Sep 2023 11:51:23 PM EDT.
Package selinux-policy-targeted-38.28-1.fc38.noarch is already installed.
Package podman-5:4.6.2-3.fc38.x86_64 is already installed.
Package systemd-253.10-1.fc38.x86_64 is already installed.
Package bluechi-agent-0.6.0-0.202309271249.git5cd0b2c.fc38.x86_64 is already installed.
Package procps-ng-3.3.17-11.fc38.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
Last metadata expiration check: 0:26:21 ago on Wed 27 Sep 2023 11:51:23 PM EDT.
Dependencies resolved.
Nothing to do.
Complete!
cp: cannot create regular file '/usr/lib/qm/rootfs/etc/containers/': Not a directory
```

PR #224, opened 2023-09-28
setup: make sure create rootfs req dirs

dougsland:
Fixes: https://github.com/containers/qm/issues/223

### DIFF
--- a/setup
+++ b/setup
@@ -20,6 +20,9 @@ CMDLINE_ARGUMENT_LIST=(
   "installdir"
   "rootfs"
   "hostname"
+  "rwetcfs"
+  "rwvarfs"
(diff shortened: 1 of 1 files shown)


dougsland:
@rhatdan PTAL

Yarboa (changes requested):
Test is failing here no QM
I do not understand this fix, but it seems QM is not up, I will check it also

rhatdan:
LGTM

dougsland:
> Test is failing here no QM I do not understand this fix,
> 
> What is the purpose of ignore systemctl but it seems QM is not up, I will check it also

This is just an issue I found in one my tests but also a "safe programming" IMHO.

Issue #225, opened 2023-09-28
setup: add args: rwetcfs rwvarfs skip-systemctl to CMDLINE_ARGUMENT_LIST


dougsland:
- rwetcfs rwvarfs are not recognized by --rwetcfs and --rwvarfs
- skip-systemctl is not using the CMD_LINE_ARGUMENT_LIST

PR #229, opened 2023-09-28
Update nodeX container file

### LINKED ISSUE #226: e2e test is broken
Please refer recent gating results 
Error: no container with name or ID "qm" found: no such container
 
https://artifacts.dev.testing-farm.io/beb520f1-c367-4e13-a572-371f65e7c317/

Yarboa:
Based on latest changes
#225 
This resolves #226

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -20,7 +20,6 @@ TasksMax=50%
 
 [Container]
 AddCapability=all
-DropCapability=sys_resource
 AddDevice=-/dev/kvm
 AddDevice=-/dev/fuse
 ContainerName=qm
@@ -29,7 +28,6 @@ Network=host
 PodmanArgs=--pids-limit=-1 --security-opt label=nested --security-opt unmask=all
 ReadOnly=true
 Rootfs=${ROOTFS}
-SecurityLabelNested=true
 SecurityLabelFileType=qm_file_t
 SecurityLabelLevel=s0
 SecurityLabelType=qm_t
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -101,7 +101,7 @@ create_qm_node() {
         echo "RUN dnf install qm -y &> /dev/null" >> ContainerFile.node"${nodeID}"
 
         # Execute qm setup
-        echo "RUN /usr/share/qm/setup 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
+        echo "RUN /usr/share/qm/setup --skip-systemctl true 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
 
         # HACK: add /dev/fuse to qm.container. It will allow a container set limits inside
         # another container.


Yarboa:
Running 
[root@node1 ~]# /usr/libexec/podman/quadlet  --dryrun
quadlet-generator[208]: Loading source unit file /etc/containers/systemd/qm.container
quadlet-generator[208]: Error converting 'qm.container', ignoring: unsupported key 'SecurityLabelNested' in group 'Container' in /etc/containers/systemd/qm.container

Please remove, it is already in PodmanArgs

Yarboa:
> Running [root@node1 ~]# /usr/libexec/podman/quadlet --dryrun quadlet-generator[208]: Loading source unit file /etc/containers/systemd/qm.container quadlet-generator[208]: Error converting 'qm.container', ignoring: unsupported key 'SecurityLabelNested' in group 'Container' in /etc/containers/systemd/qm.container
> 
> Please remove, it is already in PodmanArgs

Done

tests/e2e/lib/container
```diff
@@ -101,7 +101,7 @@ create_qm_node() {
         echo "RUN dnf install qm -y &> /dev/null" >> ContainerFile.node"${nodeID}"
 
         # Execute qm setup
-        echo "RUN /usr/share/qm/setup 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
+        echo "RUN /usr/share/qm/setup --skip-systemctl true 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
```
dougsland, line 104: `--skip-systemctl true` shouldn't be required.

dougsland:
@rhatdan PTAL

qm.container
```diff
@@ -20,7 +20,6 @@ TasksMax=50%
 
 [Container]
 AddCapability=all
-DropCapability=sys_resource
```
dougsland, line 23: Is this the dup thing you found?

Yarboa: No, it is not, it prevent from running the ffi-tools nested inside qm

tests/e2e/lib/container
Yarboa: With out the build is of qm is failing

dougsland: for me work without this flag.

qm.container
dougsland: How about instead of removing commenting these options and add a comment why this is commended?

```
# DropCapability if enabled prevent ffi-tools nested inside qm.
# DropCapability=sys_resource
```

This might help us in the future. ;-)

qm.container
```diff
@@ -29,7 +28,6 @@ Network=host
 PodmanArgs=--pids-limit=-1 --security-opt label=nested --security-opt unmask=all
 ReadOnly=true
 Rootfs=${ROOTFS}
-SecurityLabelNested=true
```
dougsland, line 32: Can we comment this one and add a note about why it's commented?

tests/e2e/lib/container
Yarboa: With or without it, I get 
```
Sep 29 05:38:01 node1 systemd[1]: qm.service: Start request repeated too quickly.
Sep 29 05:38:01 node1 systemd[1]: qm.service: Failed with result 'exit-code'.
```

Yarboa: Also running the following 

```
/usr/bin/podman run --name=qm --cidfile=%t/%N.cid --replace --rm --cgroups=split --tz=local --network=host --sdnotify=conmon -d --security-opt label=type:qm_t --security-opt label=filetype:qm_file_t --security-opt label=level:s0 --device=/dev/kvm --device=/dev/fuse --cap-drop=sys_resource --cap-add=all --read-only --read-only-tmpfs=false -v /etc/qm:/etc -v /var/qm:/var --pids-limit=-1 --security-opt label=nested --security-opt unmask=all --rootfs /usr/lib/qm/rootfs bash
```

Error: creating idfile: open %t/%N.cid: no such file or directory

dougsland: please remove the `--cidfile=%t%N.cid` thing when doing manually. Maybe you need to remove one more flag as manually is a little different than using the scripts.

dougsland: in the failing, any tip from: `journal -r`, `dmesg` or `systemctl status qm.service` ?

qm.container
Yarboa: OK, ack

One more thing there is an issued with CID permission
 cidfile is placed in /run/user/ inside node1 is readonly, int it?

qm.container
Yarboa: Yes, ack

tests/e2e/lib/container
Yarboa: One more thing there is an issued with CID permission
cidfile is placed in /run/user/ inside node1 is readonly, int it?

Yarboa: tried that also,
/usr/bin/podman run  --name=qm  --replace --rm --cgroups=split --tz=local --network=host --sdnotify=conmon -d --security-opt label=type:qm_t --security-opt label=filetype:qm_file_t --security-opt label=level:s0 --device=/dev/kvm --device=/dev/fuse --cap-drop=sys_resource --cap-add=all --read-only --read-only-tmpfs=false -v /etc/qm:/etc -v /var/qm:/var --pids-limit=-1 --security-opt label=nested --security-opt unmask=all --rootfs /usr/lib/qm/rootfs /usr/lib/systemd/systemd 
Error: crun: executable file `/usr/lib/systemd/systemd` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found

thats what i got

qm.container
rhatdan: Dropping this will allow FFI Tools to surpass their defaults.

tests/e2e/lib/container
rhatdan: This means systemd was not successfully installed.

rhatdan: `--skip-systemctl true` is needed if the setup script actually attempts to interact with systemd running within the container, their is no systemd or dbus running within a podman build RUN container.

dougsland:
already merged via https://github.com/containers/qm/pull/230

PR #231, opened 2023-10-02
fix e2e

### LINKED ISSUE #226: e2e test is broken

Yarboa:
Resolves #226 

Systemd is not installed under /usr/lib/qm/rootfs
during  
```
podman build   --runtime-flag debug  -f ContainerFile.node1 -t localhost/node:latest
```
qm.service  is failing since no /sbin/init created under /usr/lib/qm/rootfs once container is running

Adding w/a after nodeX container is up, to install systemd under rootfs

### DIFF
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -36,7 +36,7 @@ create_container() {
     local container_file="${1}"
     local container_name="${2}"
     local container_tag="${3}"
-    local build_caps_add="${4}"
+    local container_add_caps="${4}"
     local image_name
 
     image_name=$(podman images \
@@ -44,10 +44,10 @@ create_container() {
     )
     if [ -z "${image_name}" ]; then
       # Execute build with tag for latest
-      if [ -z "${build_caps_add}" ]; then
+      if [ -z "${container_add_caps}" ]; then
          podman_build="podman build -q -f ${container_file} --tag ${container_tag} 1> /dev/null"
       else
-         podman_build="podman build --cap-add ${build_caps_add} -q -f ${container_file} --tag ${container_tag} 1> /dev/null"
+         podman_build="podman build --cap-add ${container_add_caps} -q -f ${container_file} --tag ${container_tag} 1> /dev/null"
       fi
       eval "$podman_build"
       if_error_exit "create_container: \
@@ -103,11 +103,9 @@ create_qm_node() {
         echo "RUN dnf install qm -y &> /dev/null" >> ContainerFile.node"${nodeID}"
 
         # Execute qm setup
+        # Use --skip-systemctl true in podman build, systemd is not running
+        # Do not set --skip-systemctl inside running container with systemd in it.
         echo "RUN /usr/share/qm/setup --skip-systemctl true 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
-
-        # HACK: add /dev/fuse to qm.container. It will allow a container set limits inside
-        # another container.
-        echo 'RUN echo "AddDevice=-/dev/fuse" >> /usr/share/containers/systemd/qm.container' >> ContainerFile.node"${nodeID}"
 
         # Enable bluechi-agent
         echo 'RUN cp /usr/share/bluechi-agent/config/*.conf /etc/bluechi/' >> ContainerFile.node"${nodeID}"
@@ -121,11 +119,13 @@ create_qm_node() {
 
         # create the container ${nodeID}
         info_message "Creating container \033[92mnode${nodeID}\033[0m [\033[92mQM mode\033[0m]"
+
+        CONTAINER_ADD_CAPS="SYS_ADMIN /tmp/"
         create_container \
                 ContainerFile.node"${nodeID}" \
                 "node${nodeID}" \
                 "node:latest" \
-                "SYS_ADMIN /tmp/"
+                "${CONTAINER_ADD_CAPS}"
 
         # qm - after the setup, reload daemon and start qm
         eval "$(podman exec \
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -36,6 +36,7 @@ create_container() {
     local container_file="${1}"
     local container_name="${2}"
     local container_tag="${3}"
+    local build_caps_add="${4}"
     local image_name
 
     image_name=$(podman images \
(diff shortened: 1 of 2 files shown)


Yarboa:
@dougsland @rhatdan 
PTAL

That is the only way to build exploded container with systemd installed under rootfs 
It is successful once container is up and running

Yarboa:
Now i see the whole install under rootfs is not working,
Container is up with systemd but bluchi-agent not installed, this is test failure cause

@rhatdan @dougsland can you advise? is related to exploded container and podman-build?

tests/e2e/lib/container
```diff
@@ -123,6 +125,7 @@ create_qm_node() {
                 ContainerFile.node"${nodeID}" \
                 "node${nodeID}" \
                 "node:latest" \
+                "SYS_ADMIN /tmp/"
```
dougsland, line 128: How about we use a global variable to this patch?

tests/e2e/lib/container
```diff
@@ -101,7 +103,7 @@ create_qm_node() {
         echo "RUN dnf install qm -y &> /dev/null" >> ContainerFile.node"${nodeID}"
 
         # Execute qm setup
-        echo "RUN /usr/share/qm/setup 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
+        echo "RUN /usr/share/qm/setup --skip-systemctl true 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
```
dougsland, line 106: What happens if user specify in the command line --skip-systemctl false?

dougsland (commented):
In general LGTM, @Yarboa have few questions

tests/e2e/lib/container
Yarboa: This parameter should be defined only at the creation of nodes not the controllers
So we parse it as parameter,

Can you explain with snippet?

tests/e2e/lib/container
```diff
@@ -101,7 +103,7 @@ create_qm_node() {
...
 
         # HACK: add /dev/fuse to qm.container. It will allow a container set limits inside
```
dougsland, line 108: btw, @Yarboa can you remove this comment? We already added /dev/fuse in the qm.container file.

tests/e2e/lib/container
dougsland: like this: https://github.com/containers/qm/blob/main/tests/e2e/run-test-e2e#L28

tests/e2e/lib/container
dougsland: @Yarboa yeah a comment on the line above explaining about --skip-systemctl is true should be enough. In 1 or 2 years we won't remember 'why' we used that.

Yarboa: ACK

tests/e2e/lib/container
Yarboa: ACK

tests/e2e/lib/container
Yarboa: ACK

dougsland:
@Yarboa 
```
        Importing GPG key 0x1448D0A3:
         Userid     : "mperina_hirte-snapshot (None) <mperina#hirte-snapshot@copr.fedorahosted.org>"
         Fingerprint: 29FC 7556 AD3D FB1B 5BF0 64E7 02F0 E32E 1448 D0A3
         From       : https://download.copr.fedorainfracloud.org/results/mperina/hirte-snapshot/pubkey.gpg
        npm notice 
        npm notice New major version of npm available! 8.19.4 -> 10.1.0
        npm notice Changelog: <https://github.com/npm/cli/releases/tag/v10.1.0>
        npm notice Run `npm install -g npm@10.1.0` to update!
        npm notice 
        Error: no container with name or ID "qm" found: no such container
        Error: no container with name or ID "qm" found: no such container
        Error: no container with name or ID "qm" found: no such container
        Error: can only create exec sessions on running containers: container state improper
```

tests/e2e/lib/container
```diff
@@ -121,11 +118,11 @@ create_qm_node() {
 
         # create the container ${nodeID}
         info_message "Creating container \033[92mnode${nodeID}\033[0m [\033[92mQM mode\033[0m]"
+        CONTAINER_ADD_CAPS="SYS_ADMIN /tmp/"
         create_container \
                 ContainerFile.node"${nodeID}" \
                 "node${nodeID}" \
-                "node:latest" \
-                "SYS_ADMIN /tmp/"
+                "node:latest"
```
dougsland, line 125: shouldn't use the CONTAINER_ADD_CAPS here?

tests/e2e/lib/container
```diff
@@ -103,12 +102,10 @@ create_qm_node() {
...
 
         # Execute qm setup
+        # Use --skip-systemctl true in podman build, systemd is not running
+        # Do not set prameter once running on running container with systemd running
         echo "RUN /usr/share/qm/setup --skip-systemctl true 2>&1 > /tmp/qm-setup.log || echo "QM setup failed, please check /tmp/qm-setup.log."" >> ContainerFile.node"${nodeID}"
 
-        # HACK: add /dev/fuse to qm.container. It will allow a container set limits inside
-        # another container.
-        echo 'RUN echo "AddDevice=-/dev/fuse" >> /usr/share/containers/systemd/qm.container' >> ContainerFile.node"${nodeID}"
-
```
dougsland, line 109: or maybe this broke the CI?

Yarboa: I thin k yes

tests/e2e/lib/container
Yarboa: ACK

tests/e2e/lib/container
Yarboa: Will run it manually, first

dougsland: I think the problem is not this but the other thing I pointed but we never know.

Yarboa: Ack

Yarboa:
@dougsland @dennisbrendel  
PTAL

tests/e2e/lib/container
```diff
@@ -103,12 +103,10 @@ create_qm_node() {
         echo "RUN dnf install qm -y &> /dev/null" >> ContainerFile.node"${nodeID}"
 
         # Execute qm setup
+        # Use --skip-systemctl true in podman build, systemd is not running
+        # Do not set prameter once running on running container with systemd running
```
dougsland, line 107: :s/prameter/parameter/g ?
Can you re-phrase this one? "once running on running container" ? 
Is this statement correct: Do not set --skip-systemctl once the container is running with systemd ?

Yarboa: Ack

dougsland:
thanks @Yarboa great fix

PR #235, opened 2023-10-09
--skip-mount arg

odra:
Adds a `--skip-mount ` options to skip mount commands since those won't work when running the script in a Containerfile.


Yarboa:
As per @rhatdan 
```
Yes you need --cap-add sys_admin
```

PR #238, opened 2023-10-09
fix using skip-systemctl with other options

odra:
Adds `skip-systemctl,` so it can work with other cli options.

### DIFF
--- a/setup
+++ b/setup
@@ -214,6 +214,7 @@ echo  "  * rwetcfs: ${RWETCFS}"
 echo  "  * rwvarfs: ${RWVARFS}"
 echo  "  * install dir: ${INSTALLDIR}"
 echo  "  * skip-systemctl: ${SYSTEMCTL_SKIP}"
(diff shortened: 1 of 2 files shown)


Yarboa (approved):
LGTM

setup
```diff
@@ -160,7 +160,7 @@ install() {
 
 # read command line arguments
 opts=$(getopt \
-  --longoptions "$(printf "help,%s:," "${CMDLINE_ARGUMENT_LIST[@]}")" \
+  --longoptions "$(printf "help,skip-systemctl,%s:," "${CMDLINE_ARGUMENT_LIST[@]}")" \
```
dougsland, line 163: Leo, could you please fix the option name here instead: https://github.com/containers/qm/blob/main/tests/e2e/run-test-e2e#L67 ?

dougsland: @odra actually is not fix the option name, we should add there.

dougsland: ok, i figure out. lets move forward what we have.

PR #242, opened 2023-10-11
bluechi-tester: initial code

dougsland:
bluechi-tester: initial code

Signed-off-by: Michael Engel <mengel@redhat.com>

### DIFF
--- a/tests/e2e/tools/FFI/bluechi-tester/.gitignore
+++ b/tests/e2e/tools/FFI/bluechi-tester/.gitignore
@@ -1,0 +1,1 @@
+bluechi-tester
--- a/tests/e2e/tools/FFI/bluechi-tester/README.md
+++ b/tests/e2e/tools/FFI/bluechi-tester/README.md
@@ -1,0 +1,44 @@
+This is a BlueChi Tester, the idea behind is try to make any kind of test to interfere or reduce the ability to work in the BlueChi controller.
+
+Before using the bluechi-tester, make sure in the bluechi controller has a node name for this specific test in the AllowedNodeNames.
+In the controller, use journald or stderr to monitor the actions this tools has being doing.
+
+Example of usage:
+
+```
+[root@node ~]# git clone https://github.com/containers/qm
+[root@node qm]# cd qm/tests/e2e/tools/FFI/bluechi-tester
+
+[root@node bluechi-tester]# go mod init bluechi
+go: creating new go.mod: module bluechi
+go: to add module requirements and sums:
+	go mod tidy
+
+[root@node bluechi-tester]# go mod tidy
+go: finding module for package github.com/godbus/dbus/v5
+go: downloading github.com/godbus/dbus v4.1.0+incompatible
+go: downloading github.com/godbus/dbus/v5 v5.1.0
+go: found github.com/godbus/dbus/v5 in github.com/godbus/dbus/v5 v5.1.0
+
+[root@node bluechi-tester]# go build bluechi-tester.go
+
+[root@node bluechi-tester]# ./bluechi-tester
+Usage:
+  bluechi-tester [flags]
+
+Flags:
+  -h, --help                help for bluechi-tester
+  -n, --nodename string     Specify a node name
+  -N, --numbersignals int   Number of signals to be sent (default 1)
+  -s, --signal string       Specify a signal name
+  -u, --url string          Specify a URL
+
+
+Usage Example:
+
+    ./bluechi-tester
+        --url tcp:host=10.90.0.4,port=842
+        --nodename NODE_AGENT_NAME
+        --signal jobdone
+        --numbersignals 500
+```
--- a/tests/e2e/tools/FFI/bluechi-tester/bluechi-tester.go
+++ b/tests/e2e/tools/FFI/bluechi-tester/bluechi-tester.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/godbus/dbus/v5"
+	"github.com/spf13/cobra"
+)
+
+// service, object and register interface for basic client connection
+const (
+	BcDusInterface = "org.eclipse.bluechi"
+	BcObjectPath   = "/org/eclipse/bluechi/internal"
+	MethodRegister = "org.eclipse.bluechi.internal.Manager.Register"
+)
+
+// object path and signals of internal agents interface
+// see https://github.com/containers/bluechi/blob/main/data/org.eclipse.bluechi.internal.Agent.xml#L81-L116
+const (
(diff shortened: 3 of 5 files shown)


rhatdan:
LGTM

rhatdan:
@Yarboa @alexlarsson PTAL

dougsland:
We now can start adding some params in the commands line like:

- Use other actions like Heartbeats, Register etc.
- Number of commands to be send (instead of static number)

etc...

sandrobonazzola:
Isn't there any other better name than `evil-agent`?

tests/e2e/tools/FFI/bluechi-evil-agent/evil-agent.go
```diff
@@ -0,0 +1,60 @@
...
+	if len(os.Args) != 4 {
+		fmt.Printf("Usage: %s <controller_ip> <controller_port> <node_name>", os.Args[0])
+		os.Exit(1)
+	}
+
+	ctrlIP := os.Args[1]
+	ctrlPort := os.Args[2]
+	nodeName := os.Args[3]
+
+	address := fmt.Sprintf("tcp:host=%s,port=%s", ctrlIP, ctrlPort)
```
alexlarsson, line 37: I recommend just taking a dbus-style address on the command-line instead. Then you can support unix sockets, etc. Like we're using with the TLS proxy.

alexlarsson:
I also dislike using evil in the name, not because it is hurting, but because it is not very descriptive. It is not in fact evil, as we're using it to help detect problems, it is good. Call it something like bluechi-testing-agent or something that descirbes why a process like this is running.

rhatdan:
bluechi-tester is fine or bluechi-stressor.

dougsland:
```
[root@node bluechi-tester]# ./bluechi-tester
Usage:
  bluechi-tester [flags]

Flags:
  -h, --help                help for bluechi-tester
  -n, --nodename string     Specify a node name
  -N, --numbersignals int   Number of signals to be sent (default 1)
  -s, --signal string       Specify a signal name
  -u, --url string          Specify a URL


Usage Example:

    ./bluechi-tester
        --url tcp:host=10.90.0.4,port=842
        --nodename NODE_AGENT_NAME
        --signal jobdone
        --numbersignals 500
```

```
./bluechi-tester --url "tcp:host=10.90.0.4,port=842" --nodename node1 --signal jobdone --numbersignals 500
Connecting to controller: tcp:host=10.90.0.4,port=842
Sending 500 jobdone signal to tcp:host=10.90.0.4,port=842
```

rhatdan:
Should this code be put into bluechi rather then qm?  Then allow qm to use the bluechi-tester to test this?

mwperina:
> Should this code be put into bluechi rather then qm? Then allow qm to use the bluechi-tester to test this?

We could do that, it just need additional care:

1. If this should be part of integration tests, then we need an RPM package for that tester
2. If we want to have RPM package, then we should have documented and added to CI Golang development tools required to build that tester

Step 1. seems OK to me, but do we really want to add dependency to yet another programming language to BlueChi? We already use C and Python, so wouldn't it be much easier from the maintenance point of view to reimplement this tester code in either C and Python?

dougsland:
> > > Should this code be put into bluechi rather then qm? Then allow qm to use the bluechi-tester to test this?
> > 
> > 
> > We could do that, it just need additional care:
> > 
> > 1. If this should be part of integration tests, then we need an RPM package for that tester
> > 2. If we want to have RPM package, then we should have documented and added to CI Golang development tools required to build that tester
> > 
> > Step 1. seems OK to me, but do we really want to add dependency to yet another programming language to BlueChi? We
> > already use C and Python, so wouldn't it be much easier from the maintenance point of view to reimplement this tester
> > code in either C and Python?
> 
> I see this as FFI tool available in bluechi repo. Like `tests/FFI/bluechi-tester` and use it in the Github-Actions for tests, no need to package it IMHO.

Now regarding python, yes, myself and Michael tried to do that but in the end golang we could put together this tool quickly. Regarding C, this initial approach was a mimic of bluechi-agent in C but this golang was much more simple to implement.

rhatdan:
I would vote to 
a: add golang support to Bluechi for testing
b: Convert from go to python

This tester belongs to bluechi, and has little relationship to QM, other then QM uses Bluechi.

engelmi:
> a: add golang support to Bluechi for testing

+1 here. 
Although I think it fits better into the BlueChi repo than QM, it doesn't really fit there either. I am ok with adding it to BlueChi for now (like @dougsland proposed in `tests/tools/bluechi-tester` or so) and as soon as we have the `eclipse-bluechi` organization move it to a dedicated repo for test toolings (not sure if there will be more, though). 

> b: Convert from go to python

It could be any language really. With python and dasbus we didn't really manage to send the signals, but that should be possible as well - we didn't spent much time on it for a POC. I'd prefer Go and a compiled tool. 

I am a bit reluctant to add a testing tool as an RPM. How would we use this tool in BlueChi internal tests? As far as I understood it will be used by external project like QM. We can easily add it to the artifacts of a GH release so that others can access it - and then its also tied to a specific version that is used in those external projects. If we rewrite it in python, it can also be provided via a PyPi package. What do you think? @dougsland @rhatdan @mwperina

mwperina:
> I am a bit reluctant to add a testing tool as an RPM. How would we use this tool in BlueChi internal tests? As far as I understood it will be used by external project like QM. We can easily add it to the artifacts of a GH release so that others can access it - and then its also tied to a specific version that is used in those external projects. If we rewrite it in python, it can also be provided via a PyPi package. What do you think? @dougsland @rhatdan @mwperina

It seems to me much easier to install for example `bluechi-test-add-ons` RPM into our existing containers and use those artificats within integration tests rather than copying some artifacts as a part of each tests and run them. Also once we will have such `bluechi-test-add-ons` available in bluechi-snapshot repo, we can easily install such package along with bluechi also as a part of qm or any other tests ...

engelmi:
> It seems to me much easier to install for example bluechi-test-add-ons RPM into our existing containers and use those artificats within integration tests rather than copying some artifacts as a part of each tests and run them. Also once we will have such bluechi-test-add-ons available in bluechi-snapshot repo, we can easily install such package along with bluechi also as a part of qm or any other tests ...

But why would we use this tool in the integration tests of BlueChi? @mwperina 
The integration tests are supposed to be black box tests, so only using public APIs to interact with it - like a real user would do. The peer-to-peer connection between agent and controller is something BlueChi internal. A `bluechi-tester` tool makes sense to me when thinking about load testing since we can hardly simulate such a load to the controller otherwise.

dougsland:
I am going to create a new PR in bluechi repo so we can keep improving and discussing there!

PR #252, opened 2023-10-18
Apply changes of bluechi configs in e2e containers

Yarboa:
e2e tier-0 started to fail probably due to updates in
Bluechi Configuration files
Updating config files in e2e tests setup

    moving from /etc/bluechi/agent.conf ->
         /etc/bluechi/agent.conf.d/
        
    moving from /etc/bluechi/controller.conf ->
         /etc/bluechi/controller.conf.d/


Yarboa:
@odra PTAL

Issue #253, opened 2023-10-24
CI gates are broken


Yarboa:
It seems that 2 things occured:

* One rpm, moved from epel repo. need to remove it
* FFI, setup script, curl default configuration from autosd container image.

PR #254, opened 2023-10-24
Fix ContainerFile.template errors

Yarboa:
Resolve 1st issue in #253 

CI started to FAIL on dnf install
Unable to find a match: tmt-report-junit, removed it for now

CI failing due to curl configuration files, which are not accessible to public.
Replace curl configuration of bluechi entities with coded default config


odra:
LGTM

PR #255, opened 2023-10-26
e2e: adjust the path for controller.conf

dougsland:
Currently, e2e is failing with timeout as the AllowedNames is not set correctly.

### DIFF
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -182,7 +182,7 @@ create_qm_node() {
     # CONTROL NODE: append QM node into /etc/bluechi/agent.conf.din the control node the new qm node name
     eval "$(podman exec "${CONTROL_CONTAINER_NAME}" \
                 sed -i '/^AllowedNodeNames=/ s/$/,'"${AllowedNodeNames}"'/' \
-                /etc/bluechi/controller.conf.d/controller.conf
+                /etc/bluechi/controller.conf
     )"
     if_error_exit "control node: unable to sed AllowedNodeName in controller.conf"
 


dougsland:
@rhatdan @Yarboa PTAL.

dougsland:
CI error will be fixed as soon the patch is merged.

Yarboa (commented):
~@dougsland ci is passing, because of the previous fix
It will fail with it, and actually is failing now~

~I assume it is a bug in BlueChi~

> ~man bluechi-controller.conf | grep -A 4 FILE
> FILES
>        Distributions provide the /usr/share/bluechi/config/controller.conf  file  which  de‐
>        fines   bluechi   configuration  defaults.  Administrators  can  copy  this  file  to
>        /etc/bluechi/controller.conf and specify their own configuration.~
> 

 ~/etc/bluechi/controller.conf  is not read by the controller
 CI was failing since, the agents were not recognized by bluechi-controller~

mwperina (approved):
LGTM

Yarboa (changes requested):
Do not forget to change the copy also
Gate tests are failing since it is not fully set
Please change this also
https://github.com/containers/qm/pull/255/files#diff-c60a4ee82a0781e7951ef4194038877d2a9f275dbc32fa70b85947428c31310eR148

And this will cause another failure for updating qm node name, let me check

dougsland:
> Do not forget to change the copy also Gate tests are failing since it is not fully set Please change this also https://github.com/containers/qm/pull/255/files#diff-c60a4ee82a0781e7951ef4194038877d2a9f275dbc32fa70b85947428c31310eR148
> 
> And this will cause another failure for updating qm node name, let me check

Just checked, the agent is using the path /etc/bluechi/agent.conf.d/agent.conf

```
   # Enable bluechi-agent
        echo 'RUN cp /usr/share/bluechi-agent/config/*.conf /etc/bluechi/agent.conf.d/' >> ContainerFile.node"${nodeID}"
        echo 'RUN sed -i -e "s/^#NodeName=/NodeName='"node${nodeID}/"\" \
                '-e "s/^#ManagerHost=/ManagerHost='"${IP_CONTROL_MACHINE}/"\" \
                ' /etc/bluechi/agent.conf.d/agent.conf' >> ContainerFile.node"${nodeID}"
        echo "RUN systemctl enable bluechi-agent &> /dev/null" >> ContainerFile.node"${nodeID}"
```

PR #261, opened 2023-11-01
e2e: always update ManagerHost

dougsland:
We had cases where ManagerHost in node{ID} is not updated with what qm container contain as ManagerHost. This happens as the ContainerFile.node can be generated days/hours/minutes before the qm container be in fact created.

### DIFF
--- a/tests/e2e/lib/container
+++ b/tests/e2e/lib/container
@@ -153,6 +153,19 @@ create_qm_node() {
         NODES_FOR_TESTING+=("${qm_node_name}")
 
         eval "$(podman exec node"${nodeID}" \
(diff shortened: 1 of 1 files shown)


Yarboa (commented):
OK LGTM, 
Added to description happens in F39

Yarboa (approved):

dougsland:
@Yarboa found a good workaround:

```
# podman rmi node --force
# podman network rm podmanDualStack -f
# ./run-test-e2e
```

Seems to work everytime!

dougsland:
the workaround still not worked in all scenarios. Re-opening this PR.

PR #265, opened 2023-11-02
e2e: make sure the containers use iptables-nft

dougsland:
We should set as default iptables-nft, not legacy.

### DIFF
--- a/setup
+++ b/setup
@@ -15,6 +15,11 @@ AGENT_HOSTNAME="$(hostname)"
 AGENTCONF="/etc/bluechi/agent.conf.d/agent.conf"
 QM_CONTAINER_IDS=1000000000:1500000000
 CONTAINER_IDS=2500000000:1500000000
+PACKAGES_TO_INSTALL="selinux-policy-targeted podman systemd bluechi-agent procps-ng iptables-nft"
+
+# RHEL kernel uses iptables-nft, not iptables-legacy, the tool should
+# make sure this package is removed
+PACKAGES_TO_REMOVE="iptables-legacy"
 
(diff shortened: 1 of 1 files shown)


tests/e2e/lib/ContainerFile.template
```diff
@@ -33,6 +33,7 @@ RUN dnf install -y \
 	gpgme-devel \
 	libseccomp-devel \
 	iptables-services \
+	iptables-nft \
```
Yarboa, line 36: If you replace both
You can skip L68,

Tried it 
```
podman exec -it control bash -c "iptables --version"
iptables v1.8.8 (nf_tables)
```

dougsland: @Yarboa what you meant replace both?

Yarboa:
```suggestion
	iptables-nft \
	iptables-services \
```

Yarboa: @dougsland ^^

tests/e2e/lib/ContainerFile.template
```diff
@@ -64,6 +65,8 @@ RUN dnf install -y \
 	net-tools \
 	iproute
 
+RUN dnf remove -y iptables-legacy
```
Yarboa, line 68:
```suggestion

```

Yarboa: You can remove it

Yarboa (commented):
Please add Signed by in your commit

PR #269, opened 2023-11-03
setup: add vim

dougsland:
add vim editor to the qm container

### DIFF
--- a/setup
+++ b/setup
@@ -15,7 +15,7 @@ AGENT_HOSTNAME="$(hostname)"
 AGENTCONF="/etc/bluechi/agent.conf.d/agent.conf"
 QM_CONTAINER_IDS=1000000000:1500000000
 CONTAINER_IDS=2500000000:1500000000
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

pypingou:
hm, I'm not sure we really want this. I doubt we want vim in cars and we can edit files from outside the qm partition if needed (`vim /usr/lib/qm/root/...`).
I agree that the CentOS Automotive SIG images aren't built using this script directly and are thus not affected by this change but I believe we still want to keep this script and the osbuild recipe as close as possible from each other.

rhatdan:
Perhaps we should have a separate PR for setup tests from those which we will ship.  I agree we probably want to limit the potential attack surface within the QM.

dougsland:
done: https://github.com/containers/qm/pull/274

pypingou:
On Mon, Nov 06, 2023 at 06:47:54AM -0800, Douglas Landgraf wrote:
> done: https://github.com/containers/qm/pull/274

Thanks!

PR #272, opened 2023-11-05
setup: add findutils to include find tool

dougsland:


pypingou:
I would have the same argument for this as for #269

PR #276, opened 2023-11-08
Fixing tier-1 tests

Yarboa:
Tier-1 test was failing when running against AutoSD container.
Test is not running in CI gate, for now 

Sleep and command change was needed to make it work.

### DIFF
--- a/plans/e2e/tier-1.fmf
+++ b/plans/e2e/tier-1.fmf
@@ -13,12 +13,13 @@ adjust:
       - name: Install rpms
         how: install
         package: podman
-      - name: Setp AutoSD env
+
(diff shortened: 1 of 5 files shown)


Yarboa:
There is an issue with repos that break e2e preparation,
Wait for repos recovery

dennisbrendel:
Hi @Yarboa , what was the actual problem and what of that commit actually fixes the problem? Do you really want to execute with `bash -x`?

Yarboa:
> Hi @Yarboa , what was the actual problem and what of that commit actually fixes the 
problem? Do you really want to execute with `bash -x`?

The -x is not mandatory, 
Updates description

Issue #278, opened 2023-11-16
Nested image pull inside QM container id failing 


Yarboa:
With podman 4.7 and up inside c9s deployment

podman exec -it qm bash
bash-5.1# podman run -it [quay.io/centos-sig-automotive/ffi-tools:latest](http://quay.io/centos-sig-automotive/ffi-tools:latest)

return the following error 
Error: writing blob: storing blob to file "/var/tmp/container_images_storage738866264/1": write /var/tmp/container_images_storage738866264/1: no space left on device

podman exec -it qm bash -c "df -kh"
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1        23G  5.1G   18G  23% /
tmpfs            64M     0   64M   0% /dev
tmpfs           887M     0  887M   0% /tmp
**/dev/vda2        23G  618M   22G   3% /var**
tmpfs           887M  140K  887M   1% /run
tmpfs           887M     0  887M   0% /run/lock
**tmpfs           887M     0  887M   0% /var/tmp**
tmpfs           355M  9.7M  346M   3% /etc/hosts
shm              63M   84K   63M   1% /dev/shm
tmpfs           887M  8.0M  879M   1% /var/log/journal

I can think of workarounds, Question is, what is the correct solution
cat /etc/redhat-release 
CentOS Stream release 9
Host rpms
qm-0.6.0-1.20231113152056710175.main.7.gbdc6b1f.el9.noarch
podman-4.8.0~dev-1.20231115215527013754.main.2457.ec2e533a2.el9.x86_64
Host kernel uname -r
5.14.0-383.el9.x86_64

CI FFI gate is failing due to that

Yarboa:
@rhatdan @dougsland

Yarboa:
I can think of workarounds, Question is, what is the correct solution
like this one, 
```
podman exec -it qm bash -c "mkdir -p /var/container-temp ; export TMPDIR=/var/container-temp; podman info" | grep imageCopyTmpDir
  imageCopyTmpDir: /var/container-temp
```

shall we add it to QM setup?

Yarboa:
PR #280 does not resolve this one
https://artifacts.dev.testing-farm.io/9d310c51-8364-4978-a696-1dc4fb8de98d/work-ffimfngm2uf/log.txt

 Error: failed to get new shm lock manager: failed to create 2048 locks in /libpod_lock: read-only file 
 
 inside qm while pulling ffi-tools image

rhatdan:
[...]

PR #279, opened 2023-11-16
FFI: unify memory test sources

dennisbrendel:
There is no need to carry two almost identical source files.

ffi-tools container will be updated in a sec.


dennisbrendel:
https://gitlab.com/CentOS/automotive/container-images/ffi-tools/-/merge_requests/8 adapts the ffi-tools side.

tests/e2e/tools/FFI/memory/memory_eat.c
```diff
@@ -12,7 +16,7 @@ int main() {
     }
 
     unsigned long total_memory = si.totalram * si.mem_unit;
-    unsigned long memory_to_allocate = 0.2 * total_memory;
+    unsigned long memory_to_allocate = MEM_PERCENT * total_memory;
```
Yarboa, line 19: Consider here
Adding main arg parameter, that could be parsed through container env

Yarboa (commented):
CI is failing due to this #278

tests/e2e/tools/FFI/memory/memory_eat.c
dennisbrendel: Not sure I fully understand. Do you want this to be a command line parameter, so that it can be set when starting the test binary? Currently it results in the same two binaries that there were before.

dougsland: If I understood, he is looking to get the value from env variable instead.

dennisbrendel: This would be a behavioral change in the way the test is executed. If you think that makes sense then I would do it in a separate PR.

Yarboa: In 2nd thought, I think only argparse is needed with default value
of MEM_PERCENT=50 This way the program could be called as follow:

podman run --rm  --env VARIABLE1=foobar -it quay.io/centos-sig-automotive/ffi-tools:latest bash -c "./QM/90_percent_memory_eat MEM_PERCENT=30"

dennisbrendel: This implies not only having the same source file, but also having just a single identical binary.

If you want to be able to set the percentage via command line argument, then  can add this in a minute in a second PR.

Are there any concerns with merging the sources?

Yarboa: SGTM
Any way we need to move with this:
#280 in order to run ffi tests in CI, and then merge current PR
PTAL # review #280

Yarboa: > This implies not only having the same source file, but also having just a single identical binary.
> 
> If you want to be able to set the percentage via command line argument, then can add this in a minute in a second PR.
> 
> Are there any concerns with merging the sources?

@dennisbrendel I think you need to update also ffi tests, which call the binaries inside ffi-tools container

PR #281, opened 2023-11-17
e2e: remove /dev/null for creation of dual network

dougsland:


dougsland:
CI/CD issue not related.

PR #294, opened 2023-12-26
Updating old dependency in hirte-agent

### LINKED ISSUE #293: Installing qm as a stand alone result hirte-agent dependency error (AutoSD image)
Enable only 
Repo        : copr:copr.fedorainfracloud.org:rhcontainerbot:qm
Disable dnf config-manager --set-disabled autosd

dnf install -v qm
results the following error
```
Copr repo for qm owned by rhcontainerbot                                                                3.0 kB/s | 1.8 kB     00:00    
reviving: 'copr:copr.fedorainfracloud.org:rhcontainerbot:qm' can be revived - repomd matches.
copr:copr.fedorainfracloud.org:rhcontainerbot:qm: using metadata from Fri 22 Dec 2023 01:45:58 PM UTC.
repo: using cache for: baseos
baseos: using metadata from Mon 18 Dec 2023 02:58:23 PM UTC.
repo: using cache for: appstream
appstream: using metadata from Mon 18 Dec 2023 03:02:03 PM UTC.
repo: using cache for: extras-common
extras-common: using metadata from Tue 24 Oct 2023 11:42:58 AM UTC.
--> Starting dependency resolution
--> Finished dependency resolution
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides hirte-agent needed by qm-101:0.6.0-1.20231222134343832152.main.19.ge899374.el9.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:qm

```

Yarboa:
rename hirte-agent -> bluechi-agent

Resolve #293 

Adding fix related to CopyTmpImages dir was non under the right section
Adding fix for renaming /var/qm partition -> /var

Adding fix, for umount $ROOTFS/var $ROOTFS/etc it is called to earl,y thus could not create files/vars under 
$ROOTFS/var directory

### DIFF
--- a/setup
+++ b/setup
@@ -86,16 +86,15 @@ bluechiSetup() {
 NodeName=qm.${hostname}
 EOF
     fi
-    umount "${ROOTFS}/var" "${ROOTFS}/etc"
 }
 
 storage() {
     ROOTFS=$1
     if ! test -f "${ROOTFS}/etc/containers/storage.conf"; then
 	mkdir -p "${ROOTFS}/var/lib/shared/overlay-images" \
-	      "${ROOTFS}/var/lib/shared/overlay-layers" \
-              "${ROOTFS}/var/${TMP_QM_IMG_DIR}"
-	touch "${ROOTFS}/var/lib/shared/overlay-images/images.lock" \
+	      "${ROOTFS}/var/lib/shared/overlay-layers"
+
+        touch "${ROOTFS}/var/lib/shared/overlay-images/images.lock" \
 	      "${ROOTFS}/var/lib/shared/overlay-layers/layers.lock"
 
         sed -e '/additionalimage.*/,/]/s/^#//g' \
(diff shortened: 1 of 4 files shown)


setup
```diff
@@ -95,19 +95,24 @@ storage() {
...
             -e 's|^#.*transient_store.*|transient_store=true|g' \
             /"${ROOTFS}/usr/share/containers/storage.conf" \
             > "${ROOTFS}/etc/containers/storage.conf"
+    fi
 
-        if ! grep -q "env" "${ROOTFS}/etc/containers/containers.conf"; then
-            echo "env = [\"TMPDIR=/var/${TMP_QM_IMG_DIR}\"]" \
-            >>  "${ROOTFS}/etc/containers/containers.conf"
-        fi
+    #File always copied from /usr/share/qm, check if exist and append
```
dougsland, line 107: just a question, what happens if the file doesn't exists ?

Yarboa: Containers are too big, from time to the default /var/tmp
Got this error while pulling an image 
Error: writing blob: adding layer with blob "sha256:b45265b317a7897670ff015b177bac7b9d5037b3cfb490d3567da959c7e2cf71": Error processing tar file(exit status 1): write /a65be6ac39ddadfec332b73d772c49d5f1b4fffbe7a3a419d00fd58fcb4bb757/layer.tar: no space left on device

Now in this PR that we have /var partition instead of /var/qm could be that it is not needed, 
I can open a ticket to investigate it, please share your thoughts here

dougsland: It's not the first time we have complains about container too big, if it's something really hard to us. Let's reduce but on the other hand, if we have that big, we might reaching a different scenario in our tests with additional spaces/partitions etc. Yes, please open a ticket for that @Yarboa

tests/e2e/lib/tests
```diff
@@ -34,10 +34,11 @@ test_bluechi_list_all_units() {
         if [ "${CONTROL_CONTAINER_NAME}" == "host" ]; then
            bluechictl_cmd="bluechictl list-units ${node_name}"
         else
+           # It could take some time for qm bluechi-agent to connect
+           sleep 5
            bluechictl_cmd="podman exec"
            bluechictl_cmd+=" ${CONTROL_CONTAINER_NAME}"
            bluechictl_cmd+=" bluechictl list-units ${node_name}"
-           sleep 1
```
dougsland, line 40: Probably a good candidate to switch to a constant with a good name.

dougsland:
LGTM, @rhatdan WDYT?

tests/e2e/lib/tests
Yarboa: Ack

setup
Yarboa: Sure, done
#295

dougsland:
LGTM

PR #312, opened 2024-01-17
qm.container: comment DropCapability

dougsland:
There is errors running podman (OCI permission denied) limiting the environment. Comment for now.

See-Also: https://github.com/containers/crun/issues/1388

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -23,7 +23,7 @@ TasksMax=50%
 AddCapability=all
 
 # Comment DropCapability this will allow FFI Tools to surpass their defaults.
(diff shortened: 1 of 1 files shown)


dougsland:
CI/CD error not related, it just raised, we can fix soon:
```
demos/asilservice:25:1: B036 Don't except `BaseException` unless you plan to re-raise it.
```

alexlarsson:
Maybe reference https://github.com/containers/crun/issues/1388 in the commit message

rhatdan:
LGTM

Issue #301, opened 2024-01-13
run-test-e2e is creating the bluechi-agent nodes with the same name (node1)


dougsland:
run-test-e2e is creating the bluechi-agent nodes with the same name (node1). It should be node1, node2, node3, nodeN.

```
# ./run-test-e2e --number-of-nodes=3
<snip>

[root@node1 FFI]# cat /etc/bluechi/agent.conf.d/agent.conf |grep NodeName
# NodeName defaults the system's hostname.
NodeName=node1

[root@node2 ~]# cat /etc/bluechi/agent.conf.d/agent.conf | grep NodeName
# NodeName defaults the system's hostname.
NodeName=node1

[root@node3 ~]# cat /etc/bluechi/agent.conf.d/agent.conf | grep NodeName
# NodeName defaults the system's hostname.
NodeName=node1
```

dougsland:
@Yarboa removed the `good first issue` as it's a little more complex than we initially imagined.

dougsland:
See-Also: https://github.com/containers/qm/pull/317

PR #319, opened 2024-01-19
e2e: fix nodename in nodes

dougsland:
With recent changes all nodes generated had node1 as name.

Fixes: https://github.com/containers/qm/issues/301


dougsland:
@Yarboa and @lsm5 are working on CI/CD, merging as it contains a real fix.

```
No matching package to install: 'vsomeip3-selinux'
Not all dependencies satisfied
Error: Some packages could not be found.
```

Issue #299, opened 2024-01-10
Enable packit rpm build and test with epel repo dependency


Yarboa:
Based on this PR #244 
There is BuildRequires: vsomeip3-selinux >= 3.3.8
Which exist in epel repos

Need to map centos-stream+epel-next-9 -> in test part to centos-stream-9

Opened new PR for tests
#298

Yarboa:
We solved the dependency issue for now.
With distro map in packit
Currently TestingFarm is failing, checking in communities for solution work around, w/a.

PR #321, opened 2024-01-21
Fixing broken gate

Yarboa:
With introduction of vsomeip-selinux
as build dependency c9s target build is failing

resolve https://github.com/containers/qm/issues/299


dougsland:
Thanks so much @Yarboa, great job!

PR #323, opened 2024-01-23
Adding epel dependency for packit relase stage

### LINKED ISSUE #322: rpm build on commit - to main is failing on merge
Rpm build Is failing
https://github.com/containers/qm/blob/v0.6.0/.packit.yaml#L59-L61
Need to add epel dependency

Yarboa:
resolve #322 
It seems that release stage is failing due to vsomeip3-selinux

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -53,15 +53,21 @@ jobs:
   - <<: *copr
     project: qm
     targets:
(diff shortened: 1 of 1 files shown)


Yarboa:
@dougsland the fix will be visible on release after merge submitted

Issue #324, opened 2024-01-24
spec: replace buildrequirement: vsomeip3-selinux to requires


dougsland:
If I understood correctly, it's not required for the build but for usage.

Yarboa:
@dougsland
The solution for that, since we would like to push to centos-stream to have conditional to build
sikilar to this
https://github.com/eclipse-bluechi/bluechi/blob/44e15cc0aafe9954f86e7c14cf008f04dac600b2/build-scripts/build-rpm.sh#L13

WITH vsomeip-3-selinux for the build

Yarboa:
@dougsland need to add build README :)

PR #325, opened 2024-01-24
spec: vsomeip3 - replace buildreq to requires

dougsland:
Fixes: https://github.com/containers/qm/issues/324


Yarboa:
@dougsland 
The solution for that, since we would like to push to centos-stream to have conditional to build
sikilar to this 
https://github.com/eclipse-bluechi/bluechi/blob/44e15cc0aafe9954f86e7c14cf008f04dac600b2/build-scripts/build-rpm.sh#L13 

WITH vsomeip-3-selinux for the build

dougsland:
already fixed by Dan W.

PR #329, opened 2024-01-25
Handle RHEL missing vsomeip packages

rhatdan:


Yarboa (commented):
I am not following 
So, do you plan to remove the dependency of vsomeip3_selinux from build?

mwperina:
We need to bump version to 0.6.2 and then this fix might be built successfully: https://github.com/containers/qm/blob/main/VERSION

Yarboa (approved):
LGTM

dougsland:
waiting CI/CD

dougsland (approved):

PR #353, opened 2024-03-27
revert to before adding shm

pengshanyu:
Revert the Makefile to before adding SharedMemory


dougsland:
@pengshanyu I thought you added it in the previous commit? Ins't the the case?

pengshanyu:
Yes, but we plan to completely restore it to before adding shm, and then implement it in ffi-tools.

dougsland (approved):

dougsland:
okay, lets make it fly.

PR #362, opened 2024-04-09
qm: add seccomp json also deny sched_setscheduler

dougsland:
Patch originally from talk with @alexlarsson

### DIFF
--- a/rpm/qm.spec
+++ b/rpm/qm.spec
@@ -64,6 +64,7 @@ BuildRequires: pkgconfig(systemd)
 BuildRequires: selinux-policy >= %_selinux_policy_version
 BuildRequires: selinux-policy-devel >= %_selinux_policy_version
 
+Requires: containers-common
 Requires: selinux-policy >= %_selinux_policy_version
 Requires(post): selinux-policy-base >= %_selinux_policy_version
 Requires(post): selinux-policy-targeted >= %_selinux_policy_version
--- a/setup
+++ b/setup
@@ -142,6 +142,34 @@ create_rootfs_required_dirs() {
 
 }
 
+create_qm_seccomp_rules() {
+    TEMP_SECCOMP=$(mktemp)
+    SECCOMP_FILE_PATH="/usr/share/containers/seccomp.json"
+    QM_PATH_SECCOMP="/usr/share/qm/seccomp.json"
+
+    if [ ! -f "${SECCOMP_FILE_PATH}" ]; then
+        rm -f "${TEMP_SECCOMP}"
+        echo "Exiting... unable to find ${SECCOMP_FILE_PATH}"
+        exit 1
+    fi
+
+    # Copying original seccomp.json
+    cp "${SECCOMP_FILE_PATH}" "${QM_PATH_SECCOMP}"
+
+    # seccomp.json can be updated, we should get it from the source and adapt it.
+    echo "- Removing sched_setscheduler() as allowed syscall"
+    # make sure create a fresh seccomp.json for QM and remove allow permission for sched_setscheduler() syscall
+    jq --tab '(.syscalls[] | select(.names[] == "sched_setscheduler" and .action == "SCMP_ACT_ALLOW") .names) |= map(select(. != "sched_setscheduler"))' "${SECCOMP_FILE_PATH}" > "${TEMP_SECCOMP}"
+
+    echo "- Adding sched_setscheduler() into the deny list"
+    # Add sched_setscheduler to the deny list
+    jq --tab '.syscalls += [{"names": ["sched_setscheduler"], "action": "SCMP_ACT_ERRNO", "args": [], "errnoRet": 1, "errno": "EPERM"}]' "${TEMP_SECCOMP}" > "${QM_PATH_SECCOMP}"
+
+    rm -f "${TEMP_SECCOMP}"
+    echo -e "\ndone"
+
+}
+
 install() {
     ROOTFS=$1
     RWETCFS=$2
@@ -203,6 +231,9 @@ eval set --"${opts}"
 
 # main()
 root_check
+
+# Create a QM seccomp.json before start the QM daemon
+create_qm_seccomp_rules
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
--- a/qm.container
+++ b/qm.container
@@ -30,7 +30,7 @@ AddDevice=-/dev/fuse
 ContainerName=qm
 Exec=/sbin/init
 Network=host
-PodmanArgs=--pids-limit=-1 --security-opt label=nested --security-opt unmask=all
+PodmanArgs=--pids-limit=-1 --security-opt seccomp=/usr/share/qm/seccomp.json --security-opt label=nested --security-opt unmask=all
 ReadOnly=true
 # FIXME: QM is failing to start podman command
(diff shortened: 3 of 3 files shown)


seccomp/seccomp.json
```diff
@@ -0,0 +1,1051 @@
+{
```
rhatdan, line 1: This file should be created from the hosts seccomp.json file to make sure it is up to date.  It could either be created at install time or at least at package build time from the latest rhel version.

dougsland: thanks @rhatdan, going to prepare a second patch!

alexlarsson:
Also i don't think this change is entirely right. By removing the ALLOW for setschedule you're triggering the default operation, which is return ENOSYS. I think the proper thing is to *move* the setschedule to the syscall list just above it that generates EPERM instead.

You can probably do this with based on the original json with some creative use of jq.

seccomp/seccomp.json
```diff
@@ -0,0 +1,1051 @@
...
+				"swapcontext",
+				"swapoff",
+				"swapon",
+				"sysfs",
+				"uselib",
+				"userfaultfd",
+				"ustat",
+				"vm86",
+				"vm86old",
+				"vmsplice"
```
alexlarsson, line 84: This is where it should go.

dougsland:
I cannot see why it's failing in the CI systems (no systemctl status from QM), I will try to reproduce locally on CentOS. 
cc @Yarboa

rpm/qm.spec
```diff
@@ -106,7 +106,11 @@ sed -i 's/^install: man all/install:/' Makefile
 %selinux_modules_install -s %{selinuxtype} $MODULES
 
 %postun
-if [ $1 -eq 0 ]; then
+if [ "$1" -eq 1 ]; then
+    # package upgraded
+    %{_datadir}/qm/seccomp/qm-seccomp
```
dougsland, line 111: mental note: move this to qm systemd unit for start/restart actions.

dougsland:
Updated the patch seems related to ExecPreStart, now I am getting the below message. 

```
Apr 14 23:54:28 donald.medogz.local setroubleshoot[48705]: SELinux is preventing /usr/lib/systemd/systemd-update-utmp from map access on the file /usr/lib/systemd/systemd.

                                                           *****  Plugin catchall_boolean (89.3 confidence) suggests   ******************

                                                           If you want to allow any process to mmap any file on system with attribute file_type.
                                                           Then you must tell SELinux about this by enabling the 'domain_can_mmap_files' boolean.
                                                           You can read 'qm_selinux' man page for more details.
                                                           Do
                                                           setsebool -P domain_can_mmap_files 1

                                                           *****  Plugin catchall (11.6 confidence) suggests   **************************

                                                           If you believe that systemd-update-utmp should be allowed map access on the systemd file by default.
                                                           Then you should report this as a bug.
                                                           You can generate a local policy module to allow this access.
                                                           Do
                                                           allow this access for now by executing:
                                                           # ausearch -c 'systemd-update-' --raw | audit2allow -M my-systemdupdate
                                                           # semodule -X 300 -i my-systemdupdate.pp

```

cc @rhatdan should we avoid this approach? I am trying to make sure in a scenario containers-common package get updated with new seccomp.json rules we also update qm in a start/restart/reboot using ExecPreStart running qm-seccomp tool, not only in the initial qm setup.

setup
```diff
@@ -269,6 +270,9 @@ case "$1" in
 	replaceIDs /etc/subuid containers   ${CONTAINER_IDS}
 	replaceIDs /etc/subgid containers   ${CONTAINER_IDS}
 
+    # Create a QM seccomp.json before start the QM daemon
+    $("${QM_SECCOMP_TOOL}")
```
Yarboa, line 274: @dougsland 
I see this 
https://github.com/containers/qm/pull/362/files#diff-055b3acf274095b1eb61f3d7e44115449460dbf49660abaed65484e771574faeR34

It will be called in here, based on podman unit, file

rhatdan:
What are the AVC's you are seeing?

setup
dougsland: I thought I already fixed that one, I will take a look again, thanks @Yarboa

dougsland: @Yarboa also added Requires containers-common thanks!!

dougsland:
@Yarboa could be a cache in the CI/CD machines?

I see: `./usr/share/qm/seccomp/qm-seccomp` or I am looking to a different path? 

```
# rpm2cpio qm-0-1.noarch.rpm | cpio -div
./usr/share/containers/systemd/qm.container
./usr/share/doc/qm
./usr/share/doc/qm/CODE-OF-CONDUCT.md
./usr/share/doc/qm/NOTICE
./usr/share/doc/qm/README.md
./usr/share/doc/qm/SECURITY.md
./usr/share/licenses/qm
./usr/share/licenses/qm/LICENSE
./usr/share/man/man8/qm.8.gz
./usr/share/man/man8/qm_selinux.8.gz
./usr/share/qm
./usr/share/qm/containers.conf
./usr/share/qm/contexts
./usr/share/qm/file_contexts
./usr/share/qm/seccomp/qm-seccomp
./usr/share/qm/setup
./usr/share/selinux
./usr/share/selinux/devel
./usr/share/selinux/devel/include
./usr/share/selinux/devel/include/services
./usr/share/selinux/devel/include/services/qm.if
./usr/share/selinux/packages
./usr/share/selinux/packages/qm.pp.bz2
164 blocks
```

dougsland:
talked with @alexlarsson , no need to do on qm systemd unit. I will redo this part of the patch and update here.

dougsland:
The solution works, without the patch I don't see the "Operation not permitted".

```
# podman exec -it qm bash
# ./test_program
Failed to set scheduler: Operation not permitted
```

The test program:
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sched.h>
#include <errno.h>
#include <string.h>

int main() {
    int pid = getpid(); 
    int policy = SCHED_FIFO;  // Desired scheduling policy
    struct sched_param param;

    // Assign the maximum priority for the SCHED_FIFO policy
    param.sched_priority = sched_get_priority_max(policy);
    if (param.sched_priority == -1) {
        fprintf(stderr, "Failed to get max priority for SCHED_FIFO: %s\n", strerror(errno));
        return EXIT_FAILURE;
    }

    // Attempt to set the scheduling policy and priority
    if (sched_setscheduler(pid, policy, &param) == -1) {
        fprintf(stderr, "Failed to set scheduler: %s\n", strerror(errno));
        return EXIT_FAILURE;
    }

    printf("Scheduler set to SCHED_FIFO with priority %d\n", param.sched_priority);
    return EXIT_SUCCESS;
}
```

@alexlarsson @rhatdan @Yarboa ready for review.

dougsland:
@rhatdan the selinux issue I shared seems related to try to start qm service without setup complete 100%. I will create an issue. Thanks!

Yarboa (approved):
@dougsland 
Did i get it right? 
setup generates qm partition seccomp.json based on host seccomp.json
Once created, qm quadlet call the qm partition seccomp.json in podman args parameters

dougsland:
> @dougsland Did i get it right? setup generates qm partition seccomp.json based on host seccomp.json Once created, qm quadlet call the qm partition seccomp.json in podman args parameters

correct.

PR #369, opened 2024-04-18
FFI: add setsysctl

dougsland:
setsysctl is simple script that uses sysctl tool to try to change several parameters in the host OS.

It must be executed inside QM environment as a container (nested) to make sure attemps to change OS level are denied. Based in this tool, we can add it too the Container Image ffi-tools and generate tests.


dougsland:
cc @ shaochennjtu

dougsland:
it's just a script for test, nothing special, merging.

PR #371, opened 2024-04-18
FFI: add tool test for execute_set_scheduler.c

dougsland:
A tool to test if set_scheduler is denied inside QM partition.


dougsland:
looks like CI is broken, it's a tool, nothing to run against here.

Issue #376, opened 2024-04-24
tool: SCHED_DEADLINE should not be set inside QM


dougsland:
Create a tool to test if it's possible to set SCHED_DEADLINE inside QM partition. It should not be possible.

See-Also: #375

PR #378, opened 2024-04-24
tools: add deny_sched_setattr

dougsland:
A tool to test if inside QM it's possible to use the deny_sched_setattr() syscall to set SCHED_DEADLINE.

Fixes: https://github.com/containers/qm/issues/376


dougsland:
the CI/CD just got timeout.

Issue #385, opened 2024-04-24
setup: split the seccomp logic into a single tool


dougsland:
due the failure in the ostree installation we must be able to run the seccomp logic as a single tool and don't depend on the setup script. We must be calling it into the setup script AND probably create a python script to call via sample-image project OR osbuild-auto project.

dougsland:
cc @Yarboa

Issue #375, opened 2024-04-23
seccomp: The use of SCHED_DEADLINE inside QM is not supported


dougsland:
According with [sched_setscheduler(2)](https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html) the SCHED_DEADLINE is only settable via [sched_setattr(2)](https://man7.org/linux/man-pages/man2/sched_setattr.2.html) which currently is allowed in seccomp. We should block it. 

See-Also: #376 #377

PR #387, opened 2024-04-25
seccomp: add new tool create-seccomp-rules

dougsland:
This patch removes the logic from setup and creates a independent tool to manage seccomp rules for QM. It should help with ostree issue.

Fixes: https://github.com/containers/qm/issues/385
Fixes: https://github.com/containers/qm/issues/375


dougsland:
Looks like this is what we need. However, need to be tested against sample-images ostree. It's late now, should be able to test it tomorrow.

dougsland:
cc @Yarboa

Yarboa (approved):
LGTM

Yarboa (approved):
Nice trick
I hope it will work for ostree build also

dougsland:
> Nice trick I hope it will work for ostree build also

worked.

PR #388, opened 2024-04-25
qm.if: add selinux rule for unix_stream_socket

dougsland:
rule for unix_stream_socket connectto


dougsland:
@rhatdan PTAL

aesteve-rh:
Related: https://github.com/containers/qm/issues/390

Thanks @dougsland !

dougsland:
@rhatdan should we close this one and not merge?

rhatdan:
We do not want this by default.

dougsland:
> We do not want this by default.

thanks Dan!

Issue #367, opened 2024-04-17
Gap discovered for installing QM in OSTREE image type, during setup


pbrilla-rh:
Currently QM is not up,  due to TZ changes in qm boot

```
/usr/bin/podman run --name=qm --replace --rm --cgroups=split --tz=local --network=host --sdnotify=conmon -d --security-opt label=type:qm_t --security-opt label=filetype:qm_file_t --security-opt label=level:s0 --device=/dev/fuse --cap-add=all --read-only -v ${RWETCFS}:/etc -v ${RWVARFS}:/var --pids-limit=-1 --security-opt label=nested --security-opt unmask=all --rootfs ${ROOTFS} /sbin/init
Error: removing /etc/localtime: read-only file system 
```
--tz=local tries to change RO file under /etc/qm

```
podman info  
Error: overlay: can't stat imageStore dir /usr/share/containers/storage: stat /usr/share/containers/storage: no such file or directory 
```
Need to apply changes similar to QM setup
`https://github.com/containers/qm/blob/main/setup#L91-L105`

dougsland:
@Yarboa  it seems similar we talked right ?

dougsland:
reproduced locally:

```
git clone https://gitlab.com/CentOS/automotive/sample-images && cd sample-images/osbuild-manifests
sample-images/osbuild-manifests# make cs9-qemu-qm-minimal-ostree.x86_64.qcow2

 ./runvm --nographics ./cs9-qemu-qm-minimal-ostree.x86_64.qcow2
```

```
[root@localhost ~]# journalctl -r -u qm
Apr 19 04:25:42 localhost.localdomain systemd[1]: Failed
 to start qm.service.
Apr 19 04:25:42 localhost.localdomain systemd[1]:
qm.service: Failed with result 'exit-code'.
Apr 19 04:25:42 localhost.localdomain systemd[1]: qm.ser
vice: Main process exited, code=exited, status=126/n/a
Apr 19 04:25:42 localhost.localdomain qm[669]: Error: removing /etc/localtime: r
ead-only file system
```

Let me start investigate this one.

dougsland:
no actions needed from @ericcurtin but adding him in the loop as he is one of the maint. of ostree project. He might have some ideas out of the box.

ericcurtin:
There's a couple of ways we could fix this, this looks like immutability doing it's job :sweat_smile: 

Questions I'd ask is do we need to do this?

`--tz=local`

[...]

PR #392, opened 2024-04-26
qm.container: replace Timezone=local

dougsland:
In recent tests with ostree + quadlet environment the QM service was not able to work correctly, podman failed to remove /etc/localtime affecting the QM service. This patch is a workaround to Timezone=local which uses Environment=TZ to set the default timezone to UTC until we work in root cause.

More information: https://github.com/containers/qm/issues/367

Fixes: https://github.com/containers/qm/issues/367


dougsland:
Okay, CI CD is not happy, goin to open a new one with a different and smarter approach.

Issue #359, opened 2024-04-08
retest: QM is failing to start run podman #297 on asil space


dougsland:
https://github.com/containers/qm/blob/0c3420a8bd336a1c43c9ee715cf6d532417275ce/tests/ffi/common/prepare.sh#L20-L21

dougsland:
@Yarboa @pbrilla-rh can we assign someone from QE to make sure this issue is gone in recent podman/OS?

PR #395, opened 2024-04-27
prepare.sh: remove FIX-ME regarding restorecon

dougsland:
I cannot reproduce this error, removing fixme.

Fixes: https://github.com/containers/qm/issues/359


dougsland:
@Yarboa as you added the comment, PTAL.

dougsland:
@Yarboa  is out. He can always re-open if needed.

Issue #389, opened 2024-04-25
docs: how to generate a qm selinux policy?


dougsland:
Got a selinux block?  How to generate output to send to developers to debug?

PR #400, opened 2024-04-27
README: add some commands for selinux help

dougsland:
Got few questions regarding selinux, getting the output of some debugging tools in hand can save time to everyone. Let's make it available to users via README.

Fixes: https://github.com/containers/qm/issues/389


dougsland:
trivial but @rhatdan PTAL

dougsland:
it's trivial info, mr Dan is busy in others projects too.

Issue #403, opened 2024-04-28
docs: create a document for the development process 


dougsland:
- As soon we have a change in the QM how to create a new rpm?
- How to test this new change and RPM against CentOS Automotive Stream distro ?

PR #404, opened 2024-04-28
docs: Add devel README

dougsland:
It's not a trivial road to join the development, let's make the path easy to everyone that would like to join forces.

Fixes: https://github.com/containers/qm/issues/403


dougsland:
This is just documentation, we can always improve on the road.

Issue #408, opened 2024-04-30
RFC: qm.container: network = host - change to private ? macvlan?


dougsland:
As discussed with @alexlarsson and @rhatdan by default users can access some host network settings via sysctl (showed below) from QM env (please note, **NOT** possible via nested container) as the default configuration is **network=host**.  

Should we change this options to private as default configuration or even macvlan in **qm.container**?  
Maybe just make a good documentation with examples how to change it and values / behaviors?  

The **network=host** was added to simplify the network layer and make it work for most of initial deployments, on top of that, users can always change the settings and adapt for their needs. 

On the other side, we have options in podman like **--ipc=host** that is kind of similar exposition of host and we believe is not appropriate. 

```
host> podman exec -it qm bash
bash-5.1# ./rung-sysctl-tests
Randomly setting sysctl parameters...
net.ipv4.ip_forward = 1
net.ipv4.conf.all.rp_filter = 0
net.ipv4.tcp_max_syn_backlog = 632
net.core.rmem_max = 25982
net.core.wmem_max = 31861
```

alexlarsson:
Lets go with private for now. Its safer as the default, and it is what podman in general uses as the default.

dougsland:
Changing to private worked as expected:

```
[root@localhost ~]# cat /usr/share/containers/systemd/qm.container | grep network -i
Network=private
[root@localhost ~]# podman exec -it qm bash
bash-5.1# podman pull alpine:latest
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Writing manifest to image destination
05455a08881ea9cf0e752bc48e61bbd71a34c029bb13df01e40e3e70e0d007bd
bash-5.1# podman images
REPOSITORY                         TAG         IMAGE ID      CREATED       SIZE
registry.fedoraproject.org/fedora  latest      e2f83b188263  32 hours ago  229 MB
docker.io/library/alpine           latest      05455a08881e  3 months ago  7.67 MB
bash-5.1# exit
exit
[root@localhost ~]# cat /etc/os-release
NAME="Automotive Stream Distribution"
VERSION="9"
ID="autosd"
ID_LIKE="rhel fedora centos"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Automotive Stream Distribution 9"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:autosd:9"
[...]

PR #409, opened 2024-04-30
qm.container: change default network option

dougsland:
Change network option to private as default option

Fixes: https://github.com/containers/qm/issues/408

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -29,7 +29,7 @@ AddDevice=-/dev/kvm
 AddDevice=-/dev/fuse
 ContainerName=qm
 Exec=/sbin/init
-Network=host
+Network=private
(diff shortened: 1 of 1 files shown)


Yarboa (commented):
Automation takes too long, need to verify manually

dougsland:
ok, looks like it's failing because of `00:00:19 fail /tests/ffi/agent-flood (on default-0) [4/4]` which PROBABLY related to the network change. It's getting a late here, @Yarboa do you mind to continue from here please? We should be "ready" to merge I guess.

dougsland:
Okay, just talked with @Yarboa and confirmed. Moving forward with this one, let's open specific issue for this one and unblock the release. Thanks @alexlarsson @rhatdan @Yarboa

PR #415, opened 2024-05-01
tests: rename README to README.md

dougsland:
Just to show automatically in the web browsers the README in the right format.


dougsland:
just renaming READMEs, nothing special, saving CPU time in CI/CD.

Issue #416, opened 2024-05-01
tests: CI/CD pipeline (tmt) - agent-flood test is failing due the change of the default Network in the qm.container


dougsland:
Related to: https://github.com/containers/qm/pull/409

Yarboa:
@dougsland need to change qm quadlet to change network=host before running
And then return it to previous state

dougsland:
> @dougsland need to change qm quadlet to change network=host before running And then return it to previous
> state

After speaking with you today I was wondering here.... In the end I think this will solve partially the problem.  
It will get back to the green state for the CI/CD for now but we need to cover tests against network=private.  
On top of that, we need to understand what's the error agent-flood is raising when network=private as well.

pengshanyu:
Hi @dougsland , I think I may not fully get your idea, so have a few questions:
```
if NETWORK == host:
        tests.append(skip_agent_flooder_as_cannot_reach_bluechi_host_in_another_network) 
```
1) Should `agent-flood` be executed against `NETWORK == host` ?

2) May I ask which test cases need to run under `NETWORK: host` and which test cases need to run under `NETWORK: private`?
3) For "we need to cover tests against `network=private`", do you mean we need to add some new test cases to test when `NETWORK == private`?

dougsland:
> Hi @dougsland , I think I may not fully get your idea, so have a few questions:
> 
> ```
> if NETWORK == host:
>         tests.append(skip_agent_flooder_as_cannot_reach_bluechi_host_in_another_network) 
> ```
> 
> 1. Should `agent-flood` be executed against `NETWORK == host` ?

You are correct. The agent-flood doesn't work well when enabled in the private mode.

> 2. May I ask which test cases need to run under `NETWORK: host` and which test cases need to run under
> `NETWORK: private`?

At this moment, the only one that needed to be skipped in host private is the agent-flood. All the rest must be equal for both.

> 3. For "we need to cover tests against `network=private`", do you mean we need to add some new test cases to
> test when `NETWORK == private`?

At this moment no, but the code must be modular enough if we need to set private or host be easy as an `if network == private` or `if network == host` 

make sense?  Please let me know if you have any additional questions.

pengshanyu:
Is the function we expect as follows (if I understand correctly):
1) Make a common function to modify the value of `NETWORK` in qm to `host` or `private`
[...]

PR #419, opened 2024-05-01
seccomp: use SYSCALLS_TO_DENY array

dougsland:
This global array can be loaded externally into tools to check if programs can be added into QM image/yaml that are not allowed or won't work into the QM partition.

In this case: "sched_setscheduler" and "sched_setattr" which QM engineers are following the Risk Assessments rules.

Projects like CentOS Automotive Stream Distribution could be take advanced in the CI/CD pipeline.

Finally, added an example in tool, how could be used in a CI/CD scenario such integration.

### DIFF
--- a/create-seccomp-rules
+++ b/create-seccomp-rules
@@ -15,6 +15,8 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 SECCOMP_CONTAINERS_FILE="/usr/share/containers/seccomp.json"
+SYSCALLS_TO_DENY=("sched_setscheduler" "sched_setattr")
+
 QM_PATH_SECCOMP="/usr/share/qm/seccomp.json"
 QM_DIR="${QM_PATH_SECCOMP%/*}"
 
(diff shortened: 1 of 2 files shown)


dougsland:
cc @alexlarsson @Yarboa 

FYI, agent-flood ci/cd is broken as expected due the Network=private change. We will address this soon.

dougsland:
CI/CD will fail due: https://github.com/containers/qm/issues/416

dougsland:
Rebuild the CentOS Automotive with this patch added in the QM repo and the here is the test documented:

The code is included in the `create-seccomp-rules`, lets see the last lines added:
`[root@localhost ~]# tail -n 30 /usr/share/qm/create-seccomp-rules`
```
...
<SNIP>
for syscall in "${SYSCALLS_TO_DENY[@]}"; do
    # Remove syscall entry from the allow list
    remove_seccomp_entry_from_allow "${syscall}" "${QM_PATH_SECCOMP}"

    # Add syscall to the deny list
    add_syscall_deny_list "${syscall}" "${QM_PATH_SECCOMP}"
done
```

Now lets prove the new version is still working:
```
[root@localhost ~]# rm -f /usr/share/qm/seccomp.json # removing the seccomp.json from the installation
[root@localhost ~]# /usr/share/qm/create-seccomp-rules # lets generate a new one
[root@localhost ~]# tail -n 30 /usr/share/qm/seccomp.json # lets see the last lines and check if include the denies lines (which include)
<SNIP>
		{
			"names": [
				"sched_setscheduler"
			],
			"action": "SCMP_ACT_ERRNO",
			"args": [],
			"errnoRet": 1,
			"errno": "EPERM"
		},
		{
			"names": [
				"sched_setattr"
			],
			"action": "SCMP_ACT_ERRNO",
			"args": [],
			"errnoRet": 1,
			"errno": "EPERM"
		}
	]
}
```

Now copy the test file [sched_setattr](https://raw.githubusercontent.com/containers/qm/main/tests/e2e/tools/FFI/deny_sched_setattr/execute_sched_setattr.c) to the virtual machine and compile it  with `gcc -o test test.c`. Later copy it to `/usr/lib/qm/rootfs/root` to use it inside QM partition.

Now let's go inside QM partition to test it:
```
podman exec -it qm bash
# cd /root
# ./test
bash-5.1# ./test
Current Scheduling Policy: SCHED_OTHER
Current Priority: 0
sched_setattr failed: Operation not permitted
```

Everything worked as expected.

Issue #430, opened 2024-05-08
RFC: create development dir - move [setup], [qm-settings-support] to development dir


dougsland:
@alexlarsson is looking for have a single solution for deployment (at least for now) and move the setup and qm-settings-support to a development dir and remove both from `rpm/qm.spec`. If users/developers would like to use it, need to clone the repo and run manually.

**Pros**:
- Single solution for deployment via sample-images to regular and ostree images.

**Possible Cons**:
- single machines or laptops cannot use anymore the setup script shipped via RPM but can use [containered approaches like the script auto-image-builder to run sample-images as container + QM](https://gitlab.com/CentOS/automotive/sample-images/-/blob/main/auto-image-builder.sh?ref_type=heads#L1)

Question raised now: What's the point of keeping the package in Fedora/CentOS if all setup/settings will be executed in sample-images project? Maybe it should be a sample-images rpm project only? 

- qm-settings-storage will be moved to development dir (qm-settings-storage was a script called after ostree and regular installations completed to adjust storage config settings). This will be converted to sample-images changes.
- make sure our CI/CD pipeline don't trust in QM rpm to install qm in the github actions
- Revisit in the future this issue in case we need to bring setup or hooks back to QM.

dougsland:
cc @Yarboa @rhatdan @pbrilla-rh

Yarboa:
Thanks @dougsland 
AFAIU, those setup scripts, are only for regular images like, running on your desktop of vm
So, I suggest to update docs with this explanation, 
This rpm scope is for AutoSD tools for building images

ericcurtin:
QM is not really a desktop tool, @alexlarsson is right, write it in a osbuild format so it's preconfigured on OSTree and non-OSTree images.

Sometimes we write scripts to be run interactively because it's useful for demos.

QM was written for Automotive, I wouldn't worry too much about use-cases outside AutoSD right now.

rhatdan:
I am fine with this change.  The script should not be installed in the final system.

pypingou:
On Thu, May 09, 2024 at 02:47:41AM -0700, Eric Curtin wrote:
>    QM is not really a desktop tool, ***@***.*** is right, write it in a
>    osbuild format so it's preconfigured on OSTree and non-OSTree images.
> 
>    Sometimes we write scripts to be run interactively because it's useful for
>    demos.
> 
>    QM was written for Automotive, I wouldn't worry too much about use-cases
>    outside AutoSD right now.

[...]

PR #431, opened 2024-05-13
setup: move setup to experimental dir

dougsland:
The experimental dir is a home to unsupported code but developers are welcome to use, test or add more code.

Setup tool is a solution to deploy QM for non automotive environment. Today the official and only solution for automotive is the ones provided via CentOS automotive sample-images project.

See: https://gitlab.com/CentOS/automotive/sample-images
See-Also: https://github.com/containers/qm/issues/430


dougsland:
Okay, as expected, we are using our own food to make CI/CD. We will need to split it in a sub-package.
How about, qm-setup subpackage which is not required in the main qm package but we require this package in the CI/CD and qm-desktop? @alexlarsson @Yarboa @rhatdan @pypingou  WDYT?

Log from CI/CD error:
```
        ./set-ffi-env-e2e: line 218: /usr/share/qm/setup: No such file or directory
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dougsland:
> Okay, as expected, we are eating our own food to make CI/CD. We will need to split it in a sub-package. How about, qm-setup subpackage which is not required in the main qm package but we require this package in the CI/CD and qm-desktop? @alexlarsson @Yarboa @rhatdan @pypingou WDYT?
> 
> Log from CI/CD error:
> 
> ```
>         ./set-ffi-env-e2e: line 218: /usr/share/qm/setup: No such file or directory
>         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

or actually remake it using osbuild.

Yarboa:
> Okay, as expected, we are eating our own food to make CI/CD. We will need to split it in a sub-package. How about, qm-setup subpackage which is not required in the main qm package but we require this package in the CI/CD and qm-desktop? @alexlarsson @Yarboa @rhatdan @pypingou WDYT?
> 
> Log from CI/CD error:
> 
> ```
>         ./set-ffi-env-e2e: line 218: /usr/share/qm/setup: No such file or directory
>         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

Thanks @dougsland 
I tend to use ENV_VAR, 

So once rpm install called, if ENV_VAR exist like ADD_SETUP=true it will install it.
Default it will not, WDYT?

rpm/qm.spec
```diff
@@ -5,7 +5,6 @@
 %global moduletype services
 %global modulenames qm
 %global seccomp_json /usr/share/%{modulenames}/seccomp.json
-%global setup_tool %{_prefix}/share/%{modulenames}/setup
```
Yarboa, line 8: I would use it here, 
%if %{include_setup} == "true"
setup_tool %{_prefix}/share/%{modulenames}/setup
%endif

And return the deleted file in Makefile

alexlarsson:
I would much prefer it if we just kept it in /usr/share/qm/setup, but well documented as not being supported in production. Having a subpackage implies much more in terms of something being supported, and it will be very complex to ever get rid of once it is in peoples rpm databases, etc.

dougsland:
> I would much prefer it if we just kept it in /usr/share/qm/setup, but well documented as not being supported in production. Having a subpackage implies much more in terms of something being supported, and it will be very complex to ever get rid of once it is in peoples rpm databases, etc.

You got it. Next patch coming soon.

rpm/qm.spec
dougsland: I will add a patch informing setup is not the official and supported way to deploy QM for automotive as @alexlarsson shared. Will be much more simple but we can always improve on the road with `if's/env vars or subpackages` thanks @Yarboa

dougsland:
Not this solution, we will go with an update

PR #442, opened 2024-05-22
Packit: Use podman-next's default repos

lsm5:
This commit switches to using podman-next's default repos so we don't have to worry about managing targets here.


lsm5:
@Yarboa @dougsland @rhatdan PTAL

dougsland (approved):
thanks @lsm5

PR #444, opened 2024-05-27
ffi: Adding qm-omm-score-adj test

### LINKED ISSUE #427: FFI: tests: add a oom_score_adj test
We would need a new test that covers the following:

- the systemd service unit file sets a OOMScoreAdjust= property value higher than the default score in the QM service (but lower than ASIL B score) and

- the correct values are set in the kernel by reading the sysfs entries

telemaco:
Adding a ffi test to validate that que qm container main process runs with oom_score_adj = 500 and the qm containers processes runs with the om_score_adj = 750 by default

resolve #427

### DIFF
--- a/tests/ffi/qm-oom-score-adj/test.sh
+++ b/tests/ffi/qm-oom-score-adj/test.sh
@@ -1,0 +1,41 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+
+. ../common/prepare.sh
+
+export QM_HOST_REGISTRY_DIR="/var/qm/lib/containers/registry"
+export QM_REGISTRY_DIR="/var/lib/containers/registry"
+
+disk_cleanup
+prepare_test
+reload_config
+
+# Copy container image registry to /var/qm/lib/containers
+image_id=$(podman images | grep quay.io/centos-sig-automotive/ffi-tools | awk -F " " '{print $3}')
+if [ ! -d "${QM_HOST_REGISTRY_DIR}" ]; then
+    exec_cmd "mkdir -p ${QM_HOST_REGISTRY_DIR}"
+    exec_cmd "podman push ${image_id} dir:${QM_HOST_REGISTRY_DIR}/tools-ffi:latest"
+fi
+
+podman exec -it qm /bin/bash -c \
+         "podman run -d --replace --name ffi-qm  dir:${QM_REGISTRY_DIR}/tools-ffi:latest \
+          /usr/bin/sleep infinity > /dev/null"
+
+
+QM_PID=$(podman inspect qm --format '{{.State.Pid}}' | tr -d '\r')
+QM_FFI_PID=$(podman exec -it qm /bin/bash -c "podman inspect ffi-qm --format '{{.State.Pid}}'" | tr -d '\r')
+
+QM_OOM_SCORE_ADJ=$(cat /proc/$QM_PID/oom_score_adj)
+QM_FFI_OOM_SCORE_ADJ=$(podman exec -it qm /bin/bash -c "cat /proc/$QM_FFI_PID/oom_score_adj" | tr -d '\r')
+
+
+# "500" is the oom_score_adj defined for the qm/qm.container.
+test "$QM_OOM_SCORE_ADJ" -eq "500"
+
+# "750" is the oom_score_adj defined in the qm/containers.conf as default value
+# for the containers that would run inside of the qm container.
+test "$QM_FFI_OOM_SCORE_ADJ" -eq "750"
+
+podman exec -it qm /bin/bash -c "podman stop ffi-qm > /dev/null"
+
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Should probably be marked as closes https://github.com/containers/qm/issues/427

telemaco:
@bilelmoussaoui yes. that's correct.

Yarboa:
LGTM, 
Checking gate failure manually

Yarboa (approved):

tests/ffi/qm-oom-score-adj/test.sh
```diff
@@ -0,0 +1,41 @@
...
+
+disk_cleanup
+prepare_test
+reload_config
+
+# Copy container image registry to /var/qm/lib/containers
+image_id=$(podman images | grep quay.io/centos-sig-automotive/ffi-tools | awk -F " " '{print $3}')
+if [ ! -d "${QM_HOST_REGISTRY_DIR}" ]; then
+    exec_cmd "mkdir -p ${QM_HOST_REGISTRY_DIR}"
+    exec_cmd "podman push ${image_id} dir:${QM_HOST_REGISTRY_DIR}/tools-ffi:latest"
```
michalskrivanek, line 18: IIUC this is 1.5GB. taht's going to take a while for no good reason. Unless there is a plan to use the same container inside QM for other test of course

dougsland: @Yarboa  don't we have the version with the `--build-arg shortbuild=yes ` available  somewhere? 

From:
https://gitlab.com/CentOS/automotive/container-images/ffi-tools/

```
Thin image:
In case of testing only, add the following --build-arg shortbuild=yes to build command
# podman build -f Containerfile --build-arg shortbuild=yes --cap-add=sys_admin -t localhost/ffi-tools:latest .
```

telemaco: @dougsland @Yarboa Is there any problem to use an alpine:latest image instead the tool-ffi:latest?

dougsland: @odra for the shortversion question.

dougsland: @telemaco I am always in favor of eating our own food but I understand your concern. I was thinking here, should we change the logic:  thin version as default and make the **fat** version as the optional?  
@odra @Yarboa do you think we can make this happen?

Yarboa: Actually, everything is ready for that 
Opened an issue,  
https://gitlab.com/CentOS/automotive/container-images/ffi-tools/-/issues/2

Will work on that
Anyway, once tests are running one after each other, unless qm reboot container persist for other tests of qm

tests/ffi/qm-oom-score-adj/test.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+podman exec -it qm /bin/bash -c \
+         "podman run -d --replace --name ffi-qm  dir:${QM_REGISTRY_DIR}/tools-ffi:latest \
+          /usr/bin/sleep infinity > /dev/null"
+
+
+QM_PID=$(podman inspect qm --format '{{.State.Pid}}' | tr -d '\r')
```
Yarboa, line 26: consider using this 
systemctl show -P MainPID qm

tests/ffi/qm-oom-score-adj/test.sh
michalskrivanek: 1.5GB is a lot better than 3.5 GB, but it still leaves a lot to be desired. I think generally the test duration would greatly benefit of some generic optimizations (limiting number of layers, cleanup of caches) - none of that is done ATM so now wonder these images are monstrous.
It takes MINUTES to copy such amount of data on weaker boards

tests/ffi/qm-oom-score-adj/test.sh
telemaco: @Yarboa `systemctl show -P MainPID qm` reports the conmon's pid not the real process' pid, the one that has the oom_score_adj applied

tests/ffi/qm-oom-score-adj/test.sh
Yarboa: @michalskrivanek you are correct,
related to (limiting number of layers, cleanup of caches)
this one was not defined yet, 
Currently we are running c9s and other SoC which are more powerfull.

I advise to open a detailed ticket for that, with tests definitions

dougsland: Will be always things to improve, lets merge this one and continue this request in parallel.
https://github.com/containers/qm/issues/447

dougsland:
@telemaco Just a question, will you integrate this one into https://gitlab.com/CentOS/automotive/container-images/ffi-tools/ or should I create a github issue for another engineer look into it? 

cc @Yarboa as I am PTO this week.

PR #450, opened 2024-05-30
add deny_set_scheduler

### LINKED ISSUE #365: FFI: create a test case for sched_setscheduler() syscall
Basically, inside QM sched_setscheduler() syscall is not allowed and we should prove it.
Transform the below steps into a test plan.

Steps:

1- Build source code:

`gcc -o test_sched_setscheduler test_sched_setscheduler.c`
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sched.h>
#include <errno.h>
#include <string.h>

int main() {
    int pid = getpid(); 
    int policy = SCHED_FIFO;  // Desired scheduling policy
    struct sched_param param;

    // Assign the maximum priority for the SCHED_FIFO policy
    param.sched_priority = sched_get_priority_max(policy);
    if (param.sched_priority == -1) {
        fprintf(stderr, "Failed to get max priority for SCHED_FIFO: %s\n", strerror(errno));
        return EXIT_FAILURE;
    }

    // Attempt to set the scheduling policy and priority
    if (sched_setscheduler(pid, policy, &param) == -1) {
        fprintf(stderr, "Failed to set scheduler: %s\n", strerror(errno));
        return EXIT_FAILURE;
    }

    printf("Scheduler set to SCHED_FIFO with priority %d\n", param.sched_priority);
    return EXIT_SUCCESS;
}
```

2- copy the binary to qm

`cp test_sched_setscheduler /usr/lib/qm/rootfs/root/`

3- Execute the test

```
# podman exec -it qm bash
bash-5.1# ./test_sched_setscheduler
Failed to set scheduler: Operation not permitted
```

Please note: the QM pull request 362 includes the patch to make sure set schedules is not allowed, to test without this restriction just test previous version or remove `create_qm_seccomp_rules` from setup and `--security-opt seccomp=/usr/share/qm/seccomp.json` from qm.container.

See also: 
https://github.com/containers/qm/pull/362#issuecomment-2058211553
https://gitlab.com/CentOS/automotive/container-images/ffi-tools/-/blob/main/Containerfile?ref_type=heads

pengshanyu:
Add tmt test case for tool "test_sched_setscheduler"  to verify that QM environment deny set_scheduler() syscall

resolve #365

### DIFF
--- a/tests/ffi/deny_set_scheduler/test.sh
+++ b/tests/ffi/deny_set_scheduler/test.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -euvx
+
+. ../common/prepare.sh
+
+disk_cleanup
+prepare_test
+reload_config
+
+podman exec -it qm /bin/bash -c \
+         'podman run -d  --replace --name ffi-qm \
+          quay.io/centos-sig-automotive/ffi-tools:latest \
+          tail -f /dev/null'
+
+return_from_setscheduler=$(podman exec -it qm /bin/bash -c \
+         'podman exec -it ffi-qm ./QM/test_sched_setscheduler')
+
+
+if [[ "${return_from_setscheduler}" =~ "Failed to set scheduler: Operation not permitted" ]]; then
+    info_message "set_scheduler() syscall denied in QM."
+else
+    info_message "set_scheduler() syscall can be executed in QM."
+    exit 1
+fi
--- a/tests/ffi/deny_set_scheduler/test.sh
+++ b/tests/ffi/deny_set_scheduler/test.sh
@@ -2,6 +2,9 @@
 
 . ../common/prepare.sh
 
+local expected_result
+expected_result="Failed to set scheduler: Operation not permitted"
+
 disk_cleanup
 prepare_test
 reload_config
@@ -14,8 +17,7 @@ podman exec -it qm /bin/bash -c \
 return_from_setscheduler=$(podman exec -it qm /bin/bash -c \
          'podman exec -it ffi-qm ./QM/test_sched_setscheduler')
 
-
-if [[ "${return_from_setscheduler}" =~ "Failed to set scheduler: Operation not permitted" ]]; then
+if [[ "${return_from_setscheduler}" =~ ${expected_result} ]]; then
     info_message "set_scheduler() syscall denied in QM."
 else
     info_message "set_scheduler() syscall can be executed in QM."
--- a/tests/ffi/deny_set_scheduler/test.sh
+++ b/tests/ffi/deny_set_scheduler/test.sh
@@ -2,7 +2,6 @@
 
 . ../common/prepare.sh
 
-local expected_result
 expected_result="Failed to set scheduler: Operation not permitted"
 
 disk_cleanup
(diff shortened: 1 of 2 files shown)


tests/ffi/deny_set_scheduler/test.sh
```diff
@@ -0,0 +1,23 @@
...
+podman exec -it qm /bin/bash -c \
+         'podman run -d  --replace --name ffi-qm \
+          quay.io/centos-sig-automotive/ffi-tools:latest \
+          tail -f /dev/null'
+
+return_from_setscheduler=$(podman exec -it qm /bin/bash -c \
+         'podman exec -it ffi-qm ./QM/test_sched_setscheduler')
+
+
+if [[ "${return_from_setscheduler}" =~ "Failed to set scheduler: Operation not permitted" ]]; then
```
dougsland, line 18: How about we short a bit:
```
if [[ ! "${return_from_setscheduler}" =~ "Failed to set scheduler: Operation not permitted" ]];
then
    error_message "set_scheduler() syscall can be executed in QM."
    exit 1
fi
```

dougsland: (need to be tested)

pengshanyu: sure, updated it

pengshanyu:
> In description always add resolve #issue_id this way once this PR is closed the issue will also be closed automatically
> 
> please explain what tests should do, Tests out puts are not necessary we have packit ffi gate for it
Thanks for the advice. Updated.

tests/ffi/deny_set_scheduler/test.sh
```diff
@@ -0,0 +1,25 @@
...
+. ../common/prepare.sh
+
+local expected_result
+expected_result="Failed to set scheduler: Operation not permitted"
+
+disk_cleanup
+prepare_test
+reload_config
+
+podman exec -it qm /bin/bash -c \
```
Yarboa, line 12: You can use
prepare_images()
run_container_in_qm()

tests/ffi/deny_set_scheduler/test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -euvx
+
+. ../common/prepare.sh
+
+local expected_result
```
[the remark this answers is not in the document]
pengshanyu: sorry for my mistake, fixed

tests/ffi/deny_set_scheduler/test.sh
pengshanyu: I remove "local"

Yarboa: Please use this to save some redundant code 
prepare_images
# this create the argument named container running as detached
run_container_in_qm ffi-qm
#now you could continue to line_16

Yarboa (approved):
LGTM

tests/ffi/deny_set_scheduler/test.sh
pengshanyu: Done, thank you.

tests/ffi/deny_set_scheduler/test.sh
dougsland: Just to confirm, we cannot use a single if here? Do we need the else statement ? (Just trying to understand the context).

PR #472, opened 2024-07-01
tests: Disable oom_score_adj on prepare_test

telemaco:
The prepare_test fuction overrides the OOMScoreAdjust value. That should be performed if necessary through systemd drop-in files.


Yarboa (approved):
LGTM

Will be removed, once this part is resolved
https://github.com/containers/qm/issues/473

Yarboa:
It seems that in disk test, 
OOMScoreAdjust=1000 is working while the default https://github.com/containers/qm/blob/main/qm.container#L13
is not triggering oom_killer

this tests is failing 
https://github.com/containers/qm/blob/main/tests/ffi/disk/test.sh

Machine crash,
I would like to review scoring mechanism

Yarboa (approved):
LGTM

PR #483, opened 2024-07-04
update agent-flood based on the change of the default Network

pengshanyu:
- updated the configuration of `/etc/qm/bluechi/agent.conf.d/agent.conf` to add  `ControllerHost `
- updated the quadlet files for testing containers to use controller host ip

So that the agent-flood test can run with qm private network
resolve https://github.com/containers/qm/issues/416

### DIFF
--- a/tests/ffi/agent-flood/test.sh
+++ b/tests/ffi/agent-flood/test.sh
@@ -27,7 +27,7 @@ After=local-fs.target
 
 [Container]
 Image=dir:/var/lib/containers/registry/tools-ffi:latest
-Exec=/root/tests/FFI/bin/bluechi-tester --url="tcp:host=localhost,port=842" \
+Exec=/root/tests/FFI/bin/bluechi-tester --url="tcp:host=${controller_host_ip},port=842" \
      --nodename=bluechi-tester-X \
      --numbersignals=11111111 \
      --signal="JobDone"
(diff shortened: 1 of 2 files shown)


tests/ffi/agent-flood/test.sh
```diff
@@ -64,15 +64,43 @@ run_test_containers(){
...
+
+    # Check if the configuration file exists
+    if [ ! -f "$qm_config_file" ]; then
+        echo "Configuration file not found: $qm_config_file"
+        exit 1
+    fi
+
+    # Extract the Network using awk
+    qm_config_file="/usr/share/containers/systemd/qm.container"
+    qm_network_mode=$(awk -F'=' '/Network=private/ { print $2 }' "$qm_config_file")
```
Yarboa, line 79: This part is problematic
AFAIK all qm.service unit files should be updated overloaded with the following drop-in qm unit file
here `/etc/containers/systemd/qm.container.d `

Please refer this 
https://github.com/containers/qm/blob/main/tests/ffi/common/prepare.sh#L29

I think we should consult @alexlarsson

tests/ffi/agent-flood/test.sh
```diff
@@ -64,15 +64,43 @@ run_test_containers(){
...
+    controller_host_ip=$(hostname -I | awk '{print $1}')
+fi
+
 #Stop QM bluechi-agent
 exec_cmd "podman exec -it qm /bin/bash -c \
          \"systemctl stop bluechi-agent\""
 
+# This is the workaround when ControllerHost is not in /etc/qm/bluechi/agent.conf.d/agent.conf
+# Add ControllerHost to /etc/qm/bluechi/agent.conf.d/agent.conf
+if test -f "/etc/qm/bluechi/agent.conf.d/agent.conf"; then
```
Yarboa, line 100: In case file does not exist. 
It should return an error, qm bc-agent will not connect

tests/ffi/agent-flood/test.sh
```diff
@@ -64,15 +64,43 @@ run_test_containers(){
     done
 }
 
+# Extract the Network mode from qm quadlet file
+get_qm_network_mode(){
+    qm_config_file="/usr/share/containers/systemd/qm.container"
+
```
Yarboa, line 70: Please use that one  instead, for dynamic inputs
```
systemctl show -p SourcePath  qm
```

tests/ffi/agent-flood/test.sh
```diff
@@ -64,15 +64,43 @@ run_test_containers(){
...
+}
+
 disk_cleanup
 prepare_test
 reload_config
 prepare_images
 
+# Assign value to ${controller_host_ip} according to qm network mode
+controller_host_ip="localhost"
+if [ "$(get_qm_network_mode)" == "private" ]; then
```
Yarboa, line 90: so what are the option here? 
private| host | .. please use `man podman run`

What happens in the else case? 
Does test exit with failure?

tests/ffi/agent-flood/test.sh
aesteve-rh: Maybe it would be fine to use `/run/systemd/generator/qm.service` instead, since by then any drop-in setting would already have been processed?

Yarboa: ok this one is not so clear for me.
Do you want to return the Network configuration? 

I would take it from the status of the service 

`systemctl show -p ExecStart qm  | cut -c11- `

Yarboa: yes, make more sense

tests/ffi/agent-flood/test.sh
pengshanyu: Hi @Yarboa , thanks for the review.
Valid values for the container's network mode are:

- bridge, 
-  network  name  or ID>
- none
- container:id
- host
- ns
- private
- slirp4netns
- pasta

May I ask what is the bluechi controller host ip corresponding to these modes?

Yarboa: Only option as per RA is network private for qm.
So the others should fail the test with error message,

tests/ffi/agent-flood/test.sh
pengshanyu: changed to `/run/systemd/generator/qm.service`

tests/ffi/agent-flood/test.sh
pengshanyu: done

tests/ffi/agent-flood/test.sh
pengshanyu: done

PR #484, opened 2024-07-08
Add SELinux rules for Wayland

aesteve-rh:
Minimal set of SELinux rules to allow running Wayland compositor in the QM container.


aesteve-rh:
cc @telemaco @martinezjavier

Yarboa (approved):
LGTM

aesteve-rh:
I was hoping @rhatdan could take a look to ensure that rules made sense.

I can still make changes in a different branch if there is some issue.

PR #490, opened 2024-07-17
qm.if: add remaining wayland rules

aesteve-rh:
Add remaining rules required to support
booting and running the wayland scenario,
with a compositor, session creation,
qm dbus management, and all required pieces.

Note that running pem_selinux for creating
the session will be blocked, as it falls back
to unconfined_u with the current rules.

### DIFF
--- a/qm.if
+++ b/qm.if
@@ -399,13 +399,19 @@ template(`qm_domain_template',`
 
 	qm_container_template($1, wayland)
 
-	allow $1_container_wayland_t $1_file_t:dir { add_name write watch };
+	allow $1_container_wayland_t $1_file_t:chr_file map;
+	allow $1_container_wayland_t $1_file_t:dir { add_name create write watch };
 	allow $1_container_wayland_t $1_file_t:file { create write };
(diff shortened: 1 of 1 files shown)


Yarboa (approved):
LGTM
Please suggest test scenario for testing

aesteve-rh:
> Please suggest test scenario for testing

I do not mind doing so, but this is a far more complex scenario than the QM-ASIL IPC. So detailing and replicating everything that is needed is going to be a big effort. I'd suggest taking a look at the scenario we developed in [autosd-demo](https://gitlab.com/roberto-majadas/autosd-demo/-/merge_requests/65). The CLI comes with an `export` subcommand to obtain a fully aib-compatible project, with all extra files and proper osbuild-manifest. That should speed up things.

aesteve-rh:
cc @telemaco

Yarboa:
> > Please suggest test scenario for testing
> 
Please create an issue even for autosd demo
We can discuss that later on in the issue, thanks

PR #492, opened 2024-07-19
Use pattern and _perms to make policy more resilient

rhatdan:
Many times AVC's come in for certain interactions, without all of the possible AVCs. Changing to use patterns and _perms macros adds common patterns for allow rules, preventing many expected AVC's.


sandrobonazzola:
`recilient` is a typo right? Other than that, LGTM.

Yarboa (approved):
LGTM
As far as i could tell :) for now

PR #493, opened 2024-07-21
Devconf update

Yarboa:
Updating lib/systemd run against non qm image
    
    This PR enables the creation of demo services on non
    containerized environment
    
    Signed-off-by: Yariv Rachmani <yrachman@redhat.com>


Yarboa:
@dougsland PTAL

PR #495, opened 2024-07-25
verify /dev/mem is not present in QM partition

pengshanyu:
I want ensure that /dev/mem is not accesible inside QM partition.

Test Design (improved based on PR review):

1. Using test expression in QM partition:

    Execute the following command:

    `podman exec qm test -e /dev/mem`
    
2. Expected Result:

    /dev/mem does not exist, the command returns a non-zero value

3. Verify whether the test passed or not:

    PASS: if the command returns a non-zero value, it means /dev/mem is not present
    FAIL:  if the command returns 0, it means /dev/mem is present

### DIFF
--- a/tests/ffi/dev_mem_not_present/PURPOSE
+++ b/tests/ffi/dev_mem_not_present/PURPOSE
@@ -1,0 +1,10 @@
+Verifies that /dev/mem is not present in QM partition.
+
+This test execute `test -e /dev/mem` in QM partition, no other input.
+    podman exec qm test -e /dev/mem
+
+Expected result:
+    out: [ INFO  ] PASS: check_dev_mem_not_present: As expected, /dev/mem is not present in QM partition.
+
+Results location:
+    /plans/e2e/ffi/report/default-0/junit.xml
--- a/tests/ffi/dev_mem_not_present/README.md
+++ b/tests/ffi/dev_mem_not_present/README.md
@@ -1,0 +1,7 @@
+# FFI - dev_mem_not_present
+
+This test is intended to confirm that /dev/mem is not present in QM partition.
+
+## This Test Set includes these tests
+
+1. Confirm that /dev/mem is not present in QM partition.
--- a/tests/ffi/dev_mem_not_present/main.fmf
+++ b/tests/ffi/dev_mem_not_present/main.fmf
@@ -1,0 +1,5 @@
+summary: Test /dev/mem is not present in QM partition.
+test: /bin/bash ./test.sh
+duration: 10m
+tag: ffi
+framework: shell
--- a/tests/ffi/dev_mem_not_present/main.fmf
+++ b/tests/ffi/dev_mem_not_present/main.fmf
@@ -3,3 +3,4 @@ test: /bin/bash ./test.sh
 duration: 10m
 tag: ffi
 framework: shell
+id: daa57e78-58da-4792-8eef-59a6b32745a2
--- a/tests/ffi/dev_mem_not_present/test.sh
+++ b/tests/ffi/dev_mem_not_present/test.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev_mem_not_present(){
+    # Check /dev/mem is not present in QM partition
+    if podman exec qm test -e /dev/mem; then
+        info_message "Found /dev/mem in QM partition: $(podman exec -t qm ls -l /dev/mem)"
+        info_message "FAIL: check_dev_mem_not_present: Check for /dev/mem in QM partition failed, it should not be present."
+        exit 1
+    else
+        info_message "PASS: check_dev_mem_not_present: As expected, /dev/mem is not present in QM partition."
+        exit 0
+    fi
+}
+
+check_dev_mem_not_present


pengshanyu:
will improve it based on the review of the design

Yarboa (approved):
LGTM

Yarboa (approved):
LGTM

tests/ffi/dev_mem_inaccessible/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+dev_mem_inaccessible(){
+    expected_result="ls: cannot access '/dev/mem': No such file or directory"
```
dougsland, line 7: why don't we just check if the command fail? Checking string output from commands are not safe IMHO.

pengshanyu:
@dougsland @Yarboa thank you so much for the review.
This PR may be closed due to changes in test scope.

pengshanyu:
Sorry for the confusion caused by [the comments](https://github.com/containers/qm/pull/495#issuecomment-2260152413)
According to the latest decision, I need to continue this work.

tests/ffi/dev_mem_inaccessible/test.sh
pengshanyu: Yes, you are right. But just checking if the command failed, I'm not sure if that's enough to tell /dev/mem doesn't exist.

tests/ffi/dev-mem_not_present/test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev-mem_not_present(){
+    expected_result="ls: cannot access '/dev/mem': No such file or directory"
+    return_from_cmd=$(podman exec -t qm bash -c "ls -l /dev/mem")
+
+    # Check /dev/mem is not present in QM partition
+    if [[ $? != "0"  && "${return_from_cmd}" == "${expected_result}"* ]]; then
```
dougsland, line 11: question, do you need more than $? != "0" ?

tests/ffi/dev-mem_not_present/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev-mem_not_present(){
+    expected_result="ls: cannot access '/dev/mem': No such file or directory"
+    return_from_cmd=$(podman exec -t qm bash -c "ls -l /dev/mem")
```
dougsland, line 8: $(podman exec..). probably should be enough..

dennisbrendel: There is no need for ls. Just use `test` or `[`. podman forwards the return value.

```
if podman exec -t qm test -f /dev/mem &>/dev/null; then
...
fi
```

dennisbrendel: Maybe rather use ` test -e` because it's a character special file

pengshanyu: @dougsland @dennisbrendel Thank you both for the review. 
I need to ask the design approver whether it would be better to use` test -e` instead of `ls` command.

tests/ffi/dev-mem_not_present/test.sh
pengshanyu: Yes, it is redundant, I will improve it.

dennisbrendel:
Just a general comment: this is not a special property of `qm`, but of all containers created by podman. So it could also be considered a smoke test in a different place

tests/ffi/dev-mem_not_present/test.sh
pengshanyu: done

tests/ffi/dev-mem_not_present/test.sh
pengshanyu: Done
@dennisbrendel thank you so much for the suggestion.

tests/ffi/dev-mem_not_present/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev-mem_not_present(){
+    # Check /dev/mem is not present in QM partition
+    if podman exec qm test -e /dev/mem &>/dev/null; then
```
dennisbrendel, line 8: Nitpicky: you don't need the output redirection, as `test` will only return something

dennisbrendel: That comment aside it looks good to me

pengshanyu: Sure, done.

tests/ffi/dev-mem_not_present/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev-mem_not_present(){
+    # Check /dev/mem is not present in QM partition
+    if podman exec qm test -e /dev/mem; then
+        info_message "Found /dev/mem in QM partition: $(podman exec -t qm ls -l /dev/mem)"
+        info_message "FAIL: check_dev-mem_not_present: Check for /dev/mem in QM partition failed, it should not present."
```
KittyMae, line 10: It should not be present

tests/ffi/dev-mem_not_present/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../../e2e/lib/utils
+
+check_dev-mem_not_present(){
```
qiyuann, line 6: Is it a convention to use dash `-` in the function name? 
I searched online, seems it's allowed in bash, but not recommended.

dennisbrendel: `-` is part of the "portable character set" [1] and thus allowed by at least POSIX [2]

[1] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01
[2] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235

tests/ffi/dev-mem_not_present/test.sh
pengshanyu: @KittyMae Thank you for your careful review.
Done

tests/ffi/dev-mem_not_present/test.sh
pengshanyu: Thank you both for the careful review.
I was trying to distinguish /dev/mem from other words, but it does look a bit weird, so I changed hyphen to low-line.

tests/ffi/dev_mem_not_present/PURPOSE
```diff
@@ -0,0 +1,10 @@
+Verifies that /dev/mem is not present in QM partition.
```
dougsland, line 1: Please use this format in a next patch, I will unblock you for now.
 @pengshanyu  Also, your commit message container several duplicate signed-off-by (SOB) next time please remove the dup lines.

Title:
    Foobar 123 123

Description:
    Foobar 123 123

Input:
    Foobar 123 123

Expected Result:
    Foobar 123123

Jira:
    Foobar 123123

pengshanyu: Hi @dougsland, sorry, I didn't notice them. Sure, I'll pay attention next time. Thanks.

dougsland: Not your fault at all, we are starting to create it ;-)

tests/ffi/dev-mem_not_present/test.sh
qiyuann:
```
3.235 Name
In the shell command language, a word consisting **solely of underscores, digits, and alphabetics** from the portable character set. The first character of a name is not a digit.
```

dennisbrendel: > ```
> 3.235 Name
> In the shell command language, a word consisting 
> **solely of underscores, digits, and alphabetics** 
> from the portable character set. 
> The first character of a name is not a digit.
> ```

I did not read carefully enough. Thanks @qiyuann

PR #497, opened 2024-07-31
ffi: qm-oom-score-adj/test.sh improvement

dougsland:
Received a request to add additional information when the value do not match for oom-score-adj. Using this opportunity to improve this test code to have a constant.


Yarboa (approved):
LGTM

PR #501, opened 2024-08-13
Attempts to access forbidden file system resource

pengshanyu:
As the namespace is limited for the QM container, there is no method for an application to actively attempt to access a resource that is not present in the file system. Therefore it must be shown that the resources present in the file system match the expectations based on the configuration.
The QM partition configuration defines a set of available filesystem resources (devices, networks, shared memory, etc). These and only these resources are expected to be present in the QM partition's file system.


pengshanyu:
Hi @alexlarsson , @ericcurtin, @dougsland, @engelmi, could you please help to review this PR? Thanks.

ericcurtin:
Not a reviewer here, but I approve, none of this is breaking existing code at least

alexlarsson (approved):

PR #505, opened 2024-08-19
Nested QM test to new fork

nsednev:
This test is about running container named fedora inside the QM container (nested).
The test checks that nested container is running with proper name and that it is being created inside the QM container.

### DIFF
--- a/tests/ffi/qm_nested/test.sh
+++ b/tests/ffi/qm_nested/test.sh
@@ -1,0 +1,62 @@
+#!/bin/bash -euvx
+
+# shellcheck disable=SC1091
+. ../common/prepare.sh
+
+# init_ffi() uses functions from common.sh to init the env before tests.
+init_ffi() {
+	disk_cleanup
+	prepare_test
+	reload_config
+}
+
+# Declare global variables and initialize to 0
+NESTED_NAME=""
+NESTED_STATUS=""
+
+# Create nested container ffi-qm inside the QM.
+# Function to create and start nested container ffi-qm, while executing podman commands from inside the QM container.
+# Extract nested container's valuable information and store it inside the global variables for later use.
+# Globals:
+#  NESTED_NAME: The nested container name in the QM container.
+#  NESTED_STATUS: The nested container status in the QM container.
+create_nested() {
+	prepare_images
+	local nested_container_name="ffi-qm"
+	run_container_in_qm "${nested_container_name}"
+	NESTED_NAME=$(podman exec -it qm bash -c "podman inspect ${nested_container_name} --format '{{.Name}}'")
+	if_error_exit "An error occured: failed to extract Name parameter of container from inside of QM."
+	echo "Container name is: ${NESTED_NAME}"
+	NESTED_STATUS=$(podman exec -it qm bash -c "podman inspect --format='{{.State.Status}}' ${nested_container_name}")
+	if_error_exit "An error occured: failed to extract State parameter of container from inside of QM."
+	echo "Container $NESTED_NAME status is: ${NESTED_STATUS}"
+}
+
+# Function to check and verify that nested container fedora was created and is running inside the QM container.
+# Globals:
+#  NESTED_NAME: The nested container name in the QM container.
+#  NESTED_STATUS: The nested container status in the QM container.
+compare_values() {
+	local expected_nested_name="ffi-qm"
+        local expected_nested_status="running"
+
+	if [ "${NESTED_NAME}" == "" ] || [ "${NESTED_STATUS}" == "" ]; then
+                echo "FAIL: Failed to run the container inside QM."
+                exit 1
+        elif [[ "${NESTED_NAME}" != *"${expected_nested_name}"* ]]; then
+                echo "FAIL: The ${expected_nested_name} container had not been created inside QM."
+                exit 1
+        elif [[ "${NESTED_STATUS}" != *"${expected_nested_status}"*  ]]; then
+                echo "FAIL: The container status is not ${expected_nested_status} inside QM."
+		exit 1
+        fi
+
+        echo "PASS: The container ${NESTED_NAME} is ${NESTED_STATUS} inside QM."
+}
+
+# Execute the functions
+init_ffi
+create_nested
+compare_values
+
+exit 0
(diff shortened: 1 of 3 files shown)


tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+# shellcheck disable=SC1091
+. ../common/prepare.sh
+
+# Declare global variables and initialize to 0
+NESTED_NAME=0
```
dougsland, line 7: As we are expecting string, how about init it as "" ? The same for NESTED_STATUS.

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Function to check and verify that nested container fedora was created and is running inside the QM container.
+# Globals:
+#  NESTED_NAME: The nested container name in the QM container.
+#  NESTED_STATUS: The nested container status in the QM container.
+compare_values() {
+        local expected_nested_name="fedora"
+        local expected_nested_status="running"
+
+        if [ "$NESTED_NAME" == 0 ] || [ "$NESTED_STATUS" == 0 ]; then
```
dougsland, line 33: As we are using "" to init the vars, we need to change to:
```
if [ -z "$NESTED_NAME" ] && [ -z "$NESTED_STATUS" ]; then
```

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
...
+                echo "FAIL: Failed to run the container inside QM."
+                exit 1
+        elif [[ $NESTED_NAME != *"$expected_nested_name"* ]];then
```
dougsland, line 36: Please use shellcheck tool against your script, I am sure it will complain about not using "" around this variable.

Example:

"$NESTED_NAME" or even better "${NESTED_NAME}".  The ${var} is used when you need to reference the variable in a context where the variable name might be followed by characters that could confuse the shell or where additional operations are needed on the variable. In other words, ${var} is always preferable to my eyes at least as it's safer.

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
...
+                echo "FAIL: The $expected_nested_name container had not been created inside QM."
+                exit 1
+        elif [[ $NESTED_STATUS != *"$expected_nested_status"*  ]];then
```
dougsland, line 39: same here

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
dougsland, line 1:
```
#!/bin/bash -ex        <----- this is kind of approach you use to exit as soon you get an error in your script (kind of printing a bunch of debug lines in the screen) OR you need to do some if validation in all commands you execute in the script it's more lines but people might prefer as it only show lines in the screen when really there is an error. 

BTW: We have a ready to go function for that. "if_error_exit", example:

<any command in bash>
if_error_exit "An error occurred"  (do a grep in qm git repo to see some examples)
```

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
...
+NESTED_NAME=0
+NESTED_STATUS=0
+
+# Create nested container fedora inside the QM.
+# Function to create and start nested container fedora, while executing podman commands from inside the QM container.
+# Extract nested container's valuable information and store it inside the global variables for later use.
+# Globals:
+#  NESTED_NAME: The nested container name in the QM container.
+#  NESTED_STATUS: The nested container status in the QM container.
+create_nested() {
```
dougsland, line 16: Looking the files we have already, in the FFI dir, looks like the `--net host` might work actually. 
How about you use the common.sh created in the ffi ? The difference is that it's using `centos-sig-automotive/ffi-tools` image instead of fedora.  (in the end, this is what @Yarboa  is looking to for, re-use logics/code): 
https://github.com/containers/qm/blob/main/tests/ffi/deny_sched_setattr/test.sh#L1

For example, add the following in the create_nested() function:
```
#  Load previous functions already creating from prepare.sh
. ../common/prepare.sh

# init_ffi() uses functions from common.sh to init the env before tests.
init_ffi() {
    disk_cleanup 
    prepare_test
    reload_config
}

# create the nested container <Add a better description>
create_nested() {
   prepare_images
   local nested_container_name="ffi-qm"
   run_container_in_qm "${nested_container_name}"
   NESTED_NAME=$(podman exec -it qm bash -c "podman inspect "${nested_container_name}" --format '{{.Name}}'")
   echo "Container name is: ${NESTED_NAME}"
   NESTED_STATUS=$(podman exec -it qm bash -c "podman inspect --format='{{.State.Status}}' ${nested_container_name}")
   echo "Container $NESTED_NAME status is: ${NESTED_STATUS}"
}
```

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,50 @@
...
+                echo "FAIL: The $expected_nested_name container had not been created inside QM."
+                exit 1
+        elif [[ $NESTED_STATUS != *"$expected_nested_status"*  ]];then
+                echo "FAIL: The container status is not $expected_nested_status inside QM."
+        fi
+
+        echo "PASS: The container $NESTED_NAME is $NESTED_STATUS inside QM."
+}
+
+# Execute the functions
```
dougsland, line 46: init_ffi

tests/ffi/qm_nested/test.sh
nsednev: I indeed used the shellcheck, no complains were received.

ShellCheck Output
$ shellcheck myscript
No issues detected!

$

nsednev:
I'm working with Juanje Ojeda to solve the ostree specific issue and this PR should be safe to merge as we have it now.

tests/ffi/qm_nested/test.sh
```diff
@@ -0,0 +1,61 @@
...
+        local expected_nested_status="running"
+
+	if [ "${NESTED_NAME}" == "" ] || [ "${NESTED_STATUS}" == "" ]; then
+                echo "FAIL: Failed to run the container inside QM."
+                exit 1
+        elif [[ "${NESTED_NAME}" != *"${expected_nested_name}"* ]]; then
+                echo "FAIL: The ${expected_nested_name} container had not been created inside QM."
+                exit 1
+        elif [[ "${NESTED_STATUS}" != *"${expected_nested_status}"*  ]]; then
+                echo "FAIL: The container status is not ${expected_nested_status} inside QM."
```
dougsland, line 50: missing exit1

Issue #506, opened 2024-08-20
An error occurred while copying the ffi-tools image to /var/qm/lib/containers


pengshanyu:
The error message is as follows:
```
test: Test QM not allow SCHED_DEADLINE be set via sched_setattr() syscall
                cmd: /bin/bash ./test.sh
                out: Trying to pull quay.io/centos-sig-automotive/ffi-tools:latest...
                out: Getting image source signatures
                out: Copying blob sha256:7110678dd356150542b86ef6f57b9495f3cf35d9b3ad171d7ea9e18bce71ed1f
                out: Copying blob sha256:e8b54c863393b7a8216a71737771b73b16a8e43ec7acf927c7faa175c255e551
                out: Copying config sha256:28b4ec420a39ed5f286c29e399a3e83de03816248d35e8e338ab5a7c501443c2
                out: Writing manifest to image destination
                out: 28b4ec420a39ed5f286c29e399a3e83de03816248d35e8e338ab5a7c501443c2
                out: Getting image source signatures
                out: Copying blob sha256:35ac5b7af462cb4f2971c43a2c0719339354d1dabed3c3470d0c71ca6cec1789
                out: Copying blob sha256:fa9a39020e334c1df2b18f1d420d6d23f775b328d7e90eed3d70ddffaadb8d8a
                out: Copying config sha256:28b4ec420a39ed5f286c29e399a3e83de03816248d35e8e338ab5a7c501443c2
                out: Writing manifest to image destination
                out: Untagged: quay.io/centos-sig-automotive/ffi-tools:latest
                out: Deleted: 28b4ec420a39ed5f286c29e399a3e83de03816248d35e8e338ab5a7c501443c2
                out: Getting image source signatures
                out: Error: writing blob: storing blob to file "/var/tmp/container_images_storage1679536461/2": write /var/tmp/container_images_storage1679536461/2: no space left on device
                out: [ FAILED ] Error: Command podman exec -it qm /bin/bash -c       "podman run -d --net host  --replace --name ffi-qm       dir:/var/lib/containers/registry/tools-ffi:latest       tail -f /dev/null" failed with exit code: 125
                out: Terminated
                out: Shared connection to localhost closed.
                00:05:34 errr /external/tests/ffi/deny_sched_setattr (on default-0) [10/20]
```

```
test: modprobe - test interference from the QM partition to host
                cmd: /bin/bash ./test.sh
                out: Trying to pull quay.io/centos-sig-automotive/ffi-tools:latest...
[...]

PR #507, opened 2024-08-21
improve prepare_images and increase duration

pengshanyu:
1. When each test case is executed, if directory `/var/qm/lib/containers/registry` exists, delete it first, then create the directory and copy the `ffi-tools` container image to it.
2. Pulling `ffi-tools` image often times out, so extend the duration.

Resolve https://github.com/containers/qm/issues/506

### DIFF
--- a/tests/ffi/agent-flood/main.fmf
+++ b/tests/ffi/agent-flood/main.fmf
@@ -1,6 +1,6 @@
 summary: Test bluechi against stress agent
 test: /bin/bash ./test.sh
-duration: 10m
+duration: 20m
 tag: ffi
 order: 70
 framework: shell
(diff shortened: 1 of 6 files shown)


Yarboa (commented):
I assume we should consider shrinking ffi-tools
I mean move more rpms into the condition, @pengshanyu, WDYT?

dougsland:
@nsednev maybe you need this patch?

pengshanyu:
> I assume we should consider shrinking ffi-tools I mean move more rpms into the condition, @pengshanyu, WDYT?
Yes, let's try to shrinking ffi-tools. I am not familiar with some packages, could you give me some guidance?

dougsland:
safe to merge, helping reduce the size. CI/CD might be get better with the patch btw. @Yarboa will provide in an Github Issue maybe? ideas for improvement for reducing the size?

PR #508, opened 2024-08-22
add support in QM image

### LINKED ISSUE #511: tests/e2e/set-ffi-env-e2e running on autosd qm-image
Currently osbuild qm qm-minimal manidests, are installed and set with running QM
The script should exit:

- on ostree target
- if qm service is running

Yarboa:
fix #511 

In case QM already runnig as a service

tests/e2e/set-ffi-env-e2e will exit 0
in case service does not exist,
it will contine
In case of ostree image script will exit

### DIFF
--- a/tests/e2e/set-ffi-env-e2e
+++ b/tests/e2e/set-ffi-env-e2e
@@ -273,6 +273,29 @@ then
     exit
 fi
 
+echo
+info_message "Checking if QM already installed"
+info_message "=============================="
+QM_STATUS="$(systemctl is-enabled qm 2>&1)"
+if [ "$QM_STATUS" == "generated" ]; then
+   if [ "$(systemctl is-active qm)" == "active" ]; then
+       info_message "QM Enabled and Active"
+       info_message "=============================="
+       exit 0
+   fi
+   if test -d /var/qm -a -d /etc/qm ; then
+      info_message "QM Enabled and not Active"
+      info_message "=============================="
+      exit 1
+   fi
+fi
+
+if stat /run/ostree-booted > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Yarboa:
@dougsland PR tested manually against QM regular image
Gate here will verify no harm on the CentOSStream

tests/e2e/set-ffi-env-e2e
```diff
@@ -273,6 +273,29 @@ then
...
+      info_message "QM Enabled and not Active"
+      info_message "=============================="
+      exit 1
+   fi
+fi
+
+if stat /run/ostree-booted > /dev/null 2>&1; then
+   info_message "script can not run on ostree image"
+   info_message "=============================="
+   exit 1
```
dougsland, line 296: Question: is it good enough to just return this message: WARNING: Script do not support ostree image. and return 0 as it's not really an error?

tests/e2e/set-ffi-env-e2e
```diff
@@ -273,6 +273,29 @@ then
     exit
 fi
 
+echo
+info_message "Checking if QM already installed"
+info_message "=============================="
+QM_STATUS="$(systemctl is-enabled qm 2>&1)"
```
dougsland, line 279: Can we use if_error_exit instead ? Seems better to get error code than output text to my eyes. WDYT?

Yarboa: Ack

tests/e2e/set-ffi-env-e2e
Yarboa: Ack

Yarboa:
add support in QM image, related issue #517 #514

PR #509, opened 2024-08-22
Patch for common.sh

nsednev:
@dougsland suggested to add this patch to prepare.sh, so people will be able to init the ffi, by calling single init_ffi function, which includes all three functions (disk_cleanup,  prepare_test, at once in their code, reload_config), instead of calling 3 of them one by one in their code.

### DIFF
--- a/tests/ffi/common/prepare.sh
+++ b/tests/ffi/common/prepare.sh
@@ -67,3 +67,9 @@ run_container_in_qm() {
       tail -f /dev/null\""
    exec_cmd "${run_ctr_in_qm}"
 }
+
+init_ffi() {
+        disk_cleanup
+        prepare_test
+        reload_config
+}


tests/ffi/common/prepare.sh
```diff
@@ -67,3 +67,9 @@ run_container_in_qm() {
       tail -f /dev/null\""
    exec_cmd "${run_ctr_in_qm}"
 }
+
+init_ffi() {
```
dougsland, line 71: In a second patch, please add a comment into this function, like docstring explaning what this it does, so people can read the code and understand what we are trying to do. :)

dougsland:
In another patch, also start using it in your test so we can show this really works.

dougsland:
safe to merge, this is just CI/CD flaky.

tests/ffi/common/prepare.sh
nsednev: Done.

PR #514, opened 2024-08-23
ffi: memory - increase duration to 50m

dougsland:
avoid timeout

### DIFF
--- a/tests/ffi/memory/main.fmf
+++ b/tests/ffi/memory/main.fmf
@@ -1,6 +1,6 @@
 summary: Test is calling systemd as stand alone test
 test: /bin/bash ./test.sh
-duration: 20m
+duration: 50m
 tag: ffi
 framework: shell
 id: fff26f33-a051-4d9e-aae2-97935e425273


Yarboa:
I think the tests should be fixed, 
We need to trigger exit status faster, no to increase duration to 50 minutes.
WDYT?

dougsland:
> I think the tests should be fixed, We need to trigger exit status faster, no to increase duration to 50 minutes. WDYT?
> 
> I would consider reducing autosd/ffi-tools image to slimmer version, to reduce image download time WDYT?

agreed.  I am testing, I was not convinced the 20m was the root cause and found other two problems.

```
[ INFO  ] PASS: qm.container oom_score_adj value == 500
./test.sh: line 39: [: cat: /proc/0/oom_score_adj: No such file or directory: integer expression expected
[ INFO  ] FAIL: qm containers oom_score_adj != 750. Current value is cat: /proc/0/oom_score_adj: No such file or directory
```

Yarboa:
There is an issue with automation
see that https://quay.io/organization/centos-sig-automotive

quay.ioquay.io
[Quay](https://quay.io/organization/centos-sig-automotive)
Quay is the best place to build, store, and distribute your containers. Public repositories are always free.
[3:40](https://redhat-internal.slack.com/archives/D057LJCLLAJ/p1724719236636919)
45 G

Need to clean

Yarboa:
Closing #508 solved it with reduction of ffi-tools container

PR #516, opened 2024-08-26
qm.if: qemu + gtk display with wayland

aesteve-rh:
These rules are collected for an scenario
where a Qemu KVM container runs with GTK
display and Wayland compositor.

### DIFF
--- a/qm.if
+++ b/qm.if
@@ -359,11 +359,16 @@ template(`qm_domain_template',`
 	files_pid_filetrans($1_container_kvm_t, $1_container_kvm_var_run_t, { dir file lnk_file sock_file })
 	allow $1_container_kvm_t $1_container_kvm_var_run_t:{file dir} mounton;
 
+	manage_files_pattern($1_container_kvm_t, $1_file_t, $1_file_t)
+	manage_sock_files_pattern($1_container_kvm_t, $1_file_t, $1_file_t)
+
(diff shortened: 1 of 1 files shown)


dougsland:
This is awesome @aesteve-rh - have you tested this one in your autosd ? 
@rhatdan WDYT?

dougsland:
@aesteve-rh if you don't have a JIRA for this one, let's tag `jira` so the system can create a JIRA ticket for you downstream so your work can be visible in all spaces. Thank you!

cc @sandrobonazzola

Yarboa (approved):
LGTM

aesteve-rh:
> have you tested this one in your autosd ?

Yes, I did a little example scenario with `autosd-demo` for this. Since we had support already for wayland, it was relatively easy to isolate the rules that concern kvm workloads. I am now working towards doing the same but with qemu+dbus display, but it is still WiP. That would require more rules, and thus, a follow-up PR.

> if you don't have a JIRA for this one

Technically, I do, but the ticket's criteria assumes full kvm support. And that will depend on the dbus display scenario.

dougsland:
LGTM, lets wait @rhatdan for the final review.

dougsland (approved):

alexlarsson (approved):
lgtm

rhatdan:
LGTM

PR #527, opened 2024-09-01
e2e: car-dashboard - use invoke

dougsland:
replace subprocess with invoke it's safer even for a test/demo only code.


Yarboa (approved):
LGTM

PR #528, opened 2024-09-01
workflow: add bandit check

dougsland:
github actions workflow to check python code using bandit.


dougsland:
Resolves: https://github.com/containers/qm/issues/529

Issue #523, opened 2024-08-29
tests/ffi/qm-oom-score-adj/test.sh is failing 


Yarboa:
During the work to work of #518 

The tests has started to fail.
Need to inspect why, I would try the following

1) Increase duration
2) each test should be next to its input detection
Instead of 
```
QM_PID=$(podman inspect qm --format '{{.State.Pid}}' | tr -d '\r')
QM_FFI_PID=$(podman exec -it qm /bin/bash -c "podman inspect ffi-qm --format '{{.State.Pid}}'" | tr -d '\r')

```

```
QM_PID=$(podman inspect qm --format '{{.State.Pid}}' | tr -d '\r')
QM_OOM_SCORE_ADJ=$(cat "/proc/$QM_PID/oom_score_adj")

QM_FFI_PID=$(podman exec -it qm /bin/bash -c "podman inspect ffi-qm --format '{{.State.Pid}}'" | tr -d '\r')

```

Yarboa:
@telemaco Can you take a look in this ?

Yarboa:
podman run --replace --name ffi-qm -it f4bae30e8eee bash
Error: OCI runtime error: crun: the requested cgroup controller `pids` is not available

Need to review the process of image_prepare

PR #531, opened 2024-09-02
oom-score: retry in case container pid not ready

dougsland:
test has being failing, let's retry to catch the
pid 10x before exiting.

Fixes: https://github.com/containers/qm/issues/523


dougsland:
CI/CD is green again, the agent-flood must be addressed in a different patch.

PR #536, opened 2024-09-05
test: replace partprobe with parted

dougsland:
partprobe requires parted, blockdev uses util-linux-core which most distros ship it in different archs.

### DIFF
--- a/tests/e2e/lib/diskutils
+++ b/tests/e2e/lib/diskutils
@@ -56,7 +56,7 @@ select_disk_to_partition(){
         fi
       fi
   done
-  partprobe
+  parted /dev/${DISK} rescan
 }
 
 create_qm_var_part() {
--- a/rpm/qm.spec
+++ b/rpm/qm.spec
@@ -58,6 +58,7 @@ BuildRequires: pkgconfig(systemd)
 BuildRequires: selinux-policy >= %_selinux_policy_version
 BuildRequires: selinux-policy-devel >= %_selinux_policy_version
 
+Requires: parted
 Requires: containers-common
 Requires: selinux-policy >= %_selinux_policy_version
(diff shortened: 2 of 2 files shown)


dougsland:
CI/CD shown: 
```
blockdev --rereadpt /dev/nvme1n1p1
blockdev: ioctl error on BLKRRPART: Invalid argument
```

Basically saying that it won't work in the specify partition, should be the whole disk. Next round will be in the whole disk.

dougsland:
@Yarboa could you please test downstream? It's all green

tests/e2e/lib/diskutils
```diff
@@ -56,7 +56,7 @@ select_disk_to_partition(){
         fi
       fi
   done
-  partprobe
+  parted /dev/${DISK} rescan
```
Yarboa, line 59: Please add a rpm -q parted
if not please install it

dougsland:
@Yarboa it's not needed to replace fdisk with parted. Only if you really wanted. parted is able to read the changes from fdisk and update the system. If this patch work, let's merge and make it happen. Later you can decide if you want to proceed with futher improvements or not.

dougsland:
Great @Yarboa the patch seems to work but CI/CD seems not happy with the oom_score_adj again? 
```
PASS: qm.container oom_score_adj value == 500
./test.sh: line 91: [[: /bin/bash: line 1: /proc/0/oom_score_adj: No such file or directory
: syntax error: operand expected (error token is "/bin/bash: line 1: /proc/0/oom_score_adj: No such file or directory
")
FAIL: qm containers oom_score_adj != 750. Current value is '/bin/bash: line 1: /proc/0/oom_score_adj: No such file or directory
'
Shared connection to 3.145.147.198 closed.
```

dougsland:
better what we have today.

PR #553, opened 2024-09-12
Adding fix in qm cleanup for disk tests

Yarboa:
During the work of post install /var/qm partition
podman rmi throws error of full diskspace on cleanup,
This error fails the test on SoC HiL

### DIFF
--- a/tests/ffi/disk/test.sh
+++ b/tests/ffi/disk/test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -euvx
+#!/bin/bash -evx
 
 # shellcheck disable=SC1091
 
@@ -36,5 +36,6 @@ fi
 
 ls -lh /root/file.lock
 
-disk_cleanup
-
+# Calling cleanup QM directorly to workaround exit code once
+# /var/qm disk is full.
+podman exec -it qm /bin/bash -c 'podman  rmi -i -f --all; echo $?'


tests/ffi/disk/test.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -euvx
+#!/bin/bash -evx
```
dougsland, line 1: I would prefer the code below if you don't mind.

Yarboa: which code below ?
https://github.com/containers/qm/pull/553#discussion_r1756694384
Do you suggest this ? 

#!/bin/bash

tests/ffi/disk/test.sh
```diff
@@ -36,5 +36,6 @@ fi
 
 ls -lh /root/file.lock
 
-disk_cleanup
-
+# Calling cleanup QM directorly to workaround exit code once
+# /var/qm disk is full.
+podman exec -it qm /bin/bash -c 'podman  rmi -i -f --all; echo $?'
```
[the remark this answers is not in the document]
dougsland: okay, let's move with your code. We need to do a release. I don't think we should merge always code in a single line, it's not very welcome to all people. We should be as clear as possible to welcome everyone. Should be like reading a clear and concise book.

dougsland (approved):

tests/ffi/disk/test.sh
dougsland: Yes but I suggest convert a single line of code to something more mature.

dougsland:
@Yarboa please open an issue and tag jira on it. We should track this as well.

PR #555, opened 2024-09-12
packit: add support to EPEL10

dougsland:
Add support to EPEL10 releases

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -116,6 +116,7 @@ jobs:
       - fedora-39
       - fedora-38
       - epel-9
+      - epel-10
 
   - job: koji_build
     trigger: commit
@@ -125,6 +126,7 @@ jobs:
       - fedora-39
       - fedora-38
       - epel-9
+      - epel-10
 
   - job: bodhi_update
     trigger: commit
@@ -134,3 +136,4 @@ jobs:
       - fedora-39
       - fedora-38
       - epel-9
+      - epel-10


.packit.yaml
```diff
@@ -134,3 +136,4 @@ jobs:
       - fedora-39
       - fedora-38
```
sandrobonazzola, line 137: worth to be done in its own patch, but you can drop fedora-38, it has gone EOL long time ago.

sandrobonazzola (approved):
LGTM

.packit.yaml
dougsland: Agreed. It's in my TODO to do so. I was even thinking in remove fedora-39 and added 41 and 42. WDYT @sandrobonazzola  too early?

sandrobonazzola: Let's wait another couple of months for that.

.packit.yaml
```diff
@@ -116,6 +116,7 @@ jobs:
...
       - epel-9
+      - epel-10
```
lsm5, line 119: I suspect packit will complain by filing an upstream issue until epel-10 is released. But feel free to try this out if you like. LGTM with that caveat.

dougsland:
CI/CD is related to the issue we are debugging in github actions:
```
Writing manifest to image destination
Error: ffi-qm container failed to start. Status: created
```

PR #557, opened 2024-09-12
github actions: more debug about cgroups failure

dougsland:
We are investigating CI/CD failure regarding cgroupsv2


dougsland:
The patch is about showing more info about cgroups, mergning for now, we are debugging it.

CI/CD is related to the issue we are debugging in github actions:
```
Writing manifest to image destination
Error: ffi-qm container failed to start. Status: created
```

PR #567, opened 2024-09-13
check_run_systemd_journal_socket: add fail case for socket

dougsland:
If host do not have the socket, let's fail right away


pengshanyu:
@dougsland Thank you for enhancing it, which I was missing.

dougsland:
> @dougsland Thank you for enhancing it, which I was missing.

nobody missing, teamwork++. If one of us miss something, we all do. There are always space for more improvements. ;-)

dougsland:
@pengshanyu agreed, I am merging. We have one cgroups issue in CI/CD that we are investigating, it's taking longer the report for the final job.

PR #568, opened 2024-09-13
qm.container: add MemoryMax and MemoryHigh

dougsland:
Add new settings for [Service]

See-Also: https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/merge_requests/102

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -8,7 +8,22 @@ CPUWeight=50
 Delegate=true
 IOWeight=50
 ManagedOOMSwap=kill
+
+# MemoryMax
+# -----------
+# Default is infinity - sets no upper bound on memory usage, allowing the
+# service to consume as much memory as the system allows, just like
+# MemoryMax=0 or omitting it.
+MemoryMax=infinity
+
+# MemoryHigh
+# -----------
+# Default is infinity - Disables the soft memory limit, meaning no throttling
+# will occur due to high memory usage, similar to MemoryHigh=0 or
+# omitting it.
+MemoryHigh=infinity
 MemorySwapMax=0
+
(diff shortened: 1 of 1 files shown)


dougsland:
@alexlarsson

dougsland:
> I'm not sure why we need this? It changes nothing?

Nothing but I am curious about why you guys added here: https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/merge_requests/102

alexlarsson:
The defaults there change nothing, but a manifest for a car can define `qm_memory_max=4G`, and then it is very useful. I don't think there are any useful values we can hardcode in qm itself though, as we know nothing about the end-user system capacity and expected load.

dougsland:
> The defaults there change nothing, but a manifest for a car can define `qm_memory_max=4G`, and then it is very useful. I > don't think there are any useful values we can hardcode in qm itself though, as we know nothing about the end-user
> system capacity and expected load.

However, it's very useful to all users have a single template from the main qm repo. Today we have auto case but tomorrow we will have more cases for sure. Based on that, I don't think we need a new release but def. merge this one. Sounds good to you too @alexlarsson ?

alexlarsson:
I'm not sure what you mean. All this PR is doing is set a few options to their default values. It will change nothing in the behavior. Do you mean that having this in the qm.container file is a kind of documentation for these?

dougsland:
> I'm not sure what you mean. All this PR is doing is set a few options to their default values. It will change nothing in the
> behavior. Do you mean that having this in the qm.container file is a kind of documentation for these?

Yes

alexlarsson:
Sure, that sounds fine then.

PR #577, opened 2024-09-18
[WIP] add in-tree support to wayland

dougsland:
Developed by Albert and Roberto, just integrating to QM upstream.

Ref:
https://aesteve-rh.github.io/alesgar-digital-cockpit/posts/wayland-in-qm/


qm-windowmanager/etc/systemd/session-active.container
```diff
@@ -0,0 +1,19 @@
...
+SecurityLabelType=qm_container_wayland_t
+Volume=/run/systemd:/run/systemd:ro
+Volume=/run/dbus/system_bus_socket:/run/dbus/system_bus_socket
+Volume=/run/user/0:/run/user/0
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Restart=always
```
aesteve-rh, line 19:
```suggestion
Restart=on-failure
```
Once the session is created, we do not need it to restart.

Actually, I am thinking that this could be perhaps be a `.service` instead of a Quadlet? We containerize workloads in the qm.container, but this is not a running workload, we just want to create the login session.

dougsland: sure, implemented as we talked.

Yarboa (commented):
@dougsland
- Is still a draft? 
- Why packit does not create for windows manager?

qm-windowmanager/etc/systemd/system/active-session.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Session Activate Service
+After=network.target dbus.service
+
+[Service]
+ExecStart=/qm_windowmanager/session-activate/qm_windowmanager_activate_session
+Environment=XDG_RUNTIME_DIR=/run/user/0
+Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/dbus/system_bus_socket
+Restart=always
+SecurityLabelType=qm_container_wayland_t
```
aesteve-rh, line 10: As this is now not a container, the label should not be required? Although then not sure if SELinux would allow it. Otherwise, we may have to put it back in a container...

qm-windowmanager/usr/lib/tmpfiles.d/etc/containers/systemd/weston_terminal.container
```diff
@@ -0,0 +1,19 @@
+[Unit]
+After=mutter.service
+Description=weston_terminal container
```
aesteve-rh, line 3: Weston terminal was just an example workload to have something printed in the screen. Otherwise, we just get a plain background. But as a package, I think we probably want an empty screen where users put their own workloads.

dougsland: thanks for the explanation, I will document that as well as we talked. :)

qm-windowmanager/usr/share/qm/mutter/ContainerFile
```diff
@@ -0,0 +1,6 @@
+# Documentation and example only, users can adapt to their need when building
+# mutter image for their own Wayland quadlet
+FROM fedora:40
```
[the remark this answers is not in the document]
dougsland: Correct.

qm-windowmanager/etc/systemd/system/active-session.service
dougsland: Yes, I will give it a try tonight to fix that. Thanks for the hint @aesteve-rh

dougsland:
> @dougsland
> 
> * Is still a draft?
> * Why packit does not create for windows manager?

Not yet, small adjusts still needed but we are almost there.

qm-windowmanager/etc/systemd/system/wayland-session.service
```diff
@@ -0,0 +1,37 @@
...
+UnsetEnvironment=TERM
+ExecStart=/bin/sleep infinity
+Restart=no
+
+# Run the session as root (required by PAMName)
+User=0
+Group=0
+
+# Set up a full user session for the user, required by Wayland.
+PAMName=login
```
aesteve-rh, line 17: This should be wayland. That is probably what is making the session start to fail.

dougsland:
Closing in favor of: https://github.com/containers/qm/pull/610

PR #581, opened 2024-09-19
qm.if: allow kvm connectto unix_stream_socket

aesteve-rh:
This rule covers the ase of a qm_container_kvm_t
container (e.g., containerized qemu) to work
with dbus display.

Also covers the usecase for vhost-user devices,
as they use unix sockets to communicate with the
VMM (that is, assuming they also use the
qm_container_kvm_t type label).


dougsland (approved):
LGTM. @rhatdan

Yarboa (approved):
LGTM

dougsland:
looks sane and @Yarboa also agree. If @rhatdan for some reason think needs more bits, we can continue in a next one.

PR #582, opened 2024-09-22
CODEOWNERS: add kleinffm for CODEOWNERS

dougsland:
Felicia has being working in docs in the entire automobile project. I would like to suggest Felicia as co-maintainer of QM project so she is able to keep her sharp eyes in the docs and in the project as well so we follow all the docs good practices and also make sure we include in our new patches documentation so in few year nobody get lost.


Yarboa (approved):
Welcome @kleinffm

kleinffm:
Thanks @dougsland and @Yarboa.

PR #575, opened 2024-09-17
Can't create ffi-qm container inside the QM partition on OStree image - no space left on device.

nsednev:
Initial code to fix the https://github.com/containers/qm/issues/574 issue.

PR #584, opened 2024-09-25
podman pull, inside qm, fails on OStree image: no space left on device 

### LINKED ISSUE #574: Can't create ffi-qm container inside the QM partition on OStree image - no space left on device.
While trying to create ffi-qm in QM partition, on OStree image, getting "no space left on device":

[md.](https://github.com/user-attachments/files/17017771/default.md)
```
`Trying to pull quay.io/centos-sig-automotive/ffi-tools:latest...
Getting image source signatures
.....
Error: writing blob: storing blob to file "/var/tmp/container_images_storage2999638705/2": write /var/tmp/container_images_storage2999638705/2: no space left on device
[ [91mFAILED[0m ] Error: Command podman exec -it qm /bin/bash -c       "podman run -d --net host  --replace --name ffi-qm       dir:/var/lib/containers/registry/tools-ffi:latest       tail -f /dev/null" failed with exit code: 125
Terminated
```

This is OStree specific issue and doesn't happens on regular image.

nsednev:
Fix: ostree images, pulling container larger than 0.9M 

Container with size > 1G is failing and resulting no disk space on podman pull
resolve #574 

The fix creates a workaround for OStree images to properly use prepare.sh.
Rebased from https://github.com/containers/qm/pull/575
Adds OStree image specific checkup to the prepare.sh and sets image_copy_tmp_dir="/var/tmp.dir", so the prepare_images section could be used also for OStree images.

### DIFF
--- a/tests/ffi/common/prepare.sh
+++ b/tests/ffi/common/prepare.sh
@@ -3,8 +3,30 @@
 . ../../e2e/lib/utils
 
 DROP_IN_DIR="/etc/containers/systemd/qm.container.d/"
+OSTREE_PATH="/run/ostree"
 export QM_HOST_REGISTRY_DIR="/var/qm/lib/containers/registry"
 export QM_REGISTRY_DIR="/var/lib/containers/registry"
+export QM_TMP_DIR="/var/tmp.dir"
+export CONTAINERS_CONF_DIR="/etc/qm/containers/containers.conf.d"
+export QM_IMAGE_TMP_DIR_CONF="qm_image_tmp_dir.conf"
+
+# Checks if the system is running under OSTree.
+# This function determines if the system is using OSTree by checking for the
+# existence of the directory specified by the OSTREE_PATH environment variable.
+# Returns:
+# 0 (success) if the OSTREE_PATH directory exists, indicating the system is
+# running under OSTree.
+# 1 (failure) if the OSTREE_PATH directory does not exist, indicating the
+# system is not running under OSTree.
+is_ostree() {
+    if [ -d "${OSTREE_PATH}" ]; then
+        echo "The system is running under OSTree."
+        return 0
+    else
+        echo "The system is not running under OSTree."
+        return 1
+    fi
+}
 
 prepare_test() {
    # Search variables for update file in qm.container
@@ -37,6 +59,25 @@ reload_config() {
 }
 
 prepare_images() {
+   # Update container image_copy_tmp_dir if the image is an OStree.
+   # Default location for storing temporary container image content. Can be overridden with the TMPDIR environment
+   # variable. If you specify "storage", then the location of the
+   # container/storage tmp directory will be used.
+   # By default image_copy_tmp_dir="/var/tmp"
+   # This is a work around and it should not be used constantly
+   # This script checks if the directory /run/ostree exists.
+   # If it does, it executes the following commands:
+   # 1. Creates the directory /var/qm/tmp.dir if it does not already exist.
+   # 2. Creates the directory /etc/qm/containers/containers.conf.d if it does not already exist.
+   # 3. Writes a configuration file qm_image_tmp_dir.conf in /etc/qm/containers/containers.conf.d
+   #    with the content specifying the temporary directory for image copying as /var/tmp.dir.
+
+   if is_ostree; then
+      exec_cmd "mkdir -p /var/qm/tmp.dir"
+      exec_cmd "mkdir -p ${CONTAINERS_CONF_DIR}"
+      exec_cmd "echo -e '[engine]\nimage_copy_tmp_dir=\"${QM_TMP_DIR}\"' > ${CONTAINERS_CONF_DIR}/${QM_IMAGE_TMP_DIR_CONF}"
+   fi
+
    exec_cmd "podman pull quay.io/centos-sig-automotive/ffi-tools:latest"
    # Copy container image registry to /var/qm/lib/containers
    image_id=$(podman images | grep quay.io/centos-sig-automotive/ffi-tools | awk -F " " '{print $3}')
--- a/docs/qm containers.conf/README.md
+++ b/docs/qm containers.conf/README.md
@@ -1,0 +1,28 @@
+# Title: How to change the variables in qm containers.conf
+
+## Description
+
+The `container.conf` file needs to be modified to allow pulling images larger than 1G from the repository on OStree images.
+
+Input:
+
+## Update container image_copy_tmp_dir if the image is an OStree
+
+1. Create /var/qm/tmp.dir or differently named directory on host.
+2. Create /etc/qm/containers/containers.conf.d or differently named directory on host.
+3. Create and edit /etc/qm/containers/containers.conf.d/qm_image_tmp_dir.conf or differently named *.conf file and add the following lines:
+[engine]
+image_copy_tmp_dir="/var/tmp.dir"
+
+## Sample code should look like this
+
+<https://github.com/nsednev/qm/blob/3bbe302791ea5d0f271a1cc96ed6bce4d4b99de2/tests/ffi/common/prepare.sh#L76-L79>
+
+By default image_copy_tmp_dir="/var/tmp".
+Changing its default value to /var/tmp.dir will allow the container to pull images larger than 1G.
+
+This is a work around and it should not be used constantly.
+
+## Expected result
+
+Containers on host will be able to pull images larger than 1G.


nsednev:
@Yarboa @dougsland Please review this fix.

aesteve-rh:
The commit itself could get a better title/body description.

nsednev:
> The commit itself could get a better title/body description.

Fixed.

aesteve-rh:
> Fixed.

I meant the commit message, not the PR. You could use a similar text though.

Also, please try to avoid creating a new commit when you rebase/merge from main, so that we can keep a clean git history.

tests/ffi/common/prepare.sh
```diff
@@ -37,6 +37,18 @@ reload_config() {
 }
 
 prepare_images() {
+   # Update container image_copy_tmp_dir if the image is an OStree.
+   # Default location for storing temporary container image content. Can be overridden with the TMPDIR environment
+   # variable. If you specify "storage", then the location of the
+   # container/storage tmp directory will be used.
+   # By default image_copy_tmp_dir="/var/tmp"
+   # This is a work around and it should not be used constantly
+   if [ -d /run/ostree ]; then
```
dougsland, line 46: The patch looks good to me but thinking ahead in time for saving time for us, I would like to recommend a few things:

1) A function to see if we are running in ostree or not.

```
OSTREE_PATH="/run/ostree" # global definition
 is_ostree() {
    if [ -d "${OSTREE_PATH}" ]; then
        echo "The system is running under OSTree."
        return 0
    else
        echo "The system is not running under OSTree."
        return 1
    fi
}       
```

2) Can you define global constants for:
      /var/tmp.dir
      /etc/qm/containers/containers.conf.d
      qm_image_tmp_dir.conf

Other than that, the patch looks clean and good, thanks so much for addressing this nasty bug.

tests/ffi/common/prepare.sh
```diff
@@ -37,6 +37,18 @@ reload_config() {
...
+      exec_cmd "mkdir -p /var/qm/tmp.dir"
```
Yarboa, line 47: Please use test -d /var/qm/tmp.dir
Assume it is already created, it will generate an error

tests/ffi/common/prepare.sh
```diff
@@ -37,6 +37,18 @@ reload_config() {
...
+      exec_cmd "mkdir -p /etc/qm/containers/containers.conf.d"
+      exec_cmd "echo -e '[engine]\nimage_copy_tmp_dir=\"/var/tmp.dir\"' > /etc/qm/containers/containers.conf.d/qm_image_tmp_dir.conf"
```
Yarboa, line 49: I would add here check id the file exist, and a value exist, WDYT?
What do you prefer to to if each of the following  exist? 

I assume that this step will done only once on the first prepare, after that, id the file exist and the parameter, just verify it is the correct one,

I would also check this directory in cleanup,
If exist, remove the files there

tests/ffi/common/prepare.sh
nsednev: Not sure I understand the suggestion. In this code if checks for the /run/ostree existance with -d, if that directory exists, then all other parts of the case implemented, otherwise skips.
Can you explain what would you like to change here?
Test is a command that is equivalent to the [ command. 
I can implement it this way:
if test -d /var/qm/tmp.dir; then
But I don't see much difference here.
However, if I'll simply use `test -d /var/qm/tmp.dir` without the if statement, it will execute the test and return an exit status (0 for true, 1 for false), but it won't execute any code based on the result.

tests/ffi/common/prepare.sh
nsednev: I can add a checkup for the directory that I'm creating beforehand and if exists, then I will need to create another random empty directory in /var/qm/ and in /etc/qm/containers/ on host.
This is a workaround, so I don't think that I should enhance this patch too much, as the image should be fixed, rather than this patch should be extended too much.
Also an image comes clean, without these directories, so chance of have the same directory existing from clean image is low.
I ran several times this code on the system with existing directory and no harm, all worked just fine, so I doubt that this enhancement will be beneficial.

tests/ffi/common/prepare.sh
Yarboa: no see what happened to the script if you run `mkdir -p <existing dir>`
so you can use test command, `man test`

Yarboa: Ack

tests/ffi/common/prepare.sh
Yarboa: OK, no need, you are correct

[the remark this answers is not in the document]
nsednev: > I was not aware of that, https://docs.podman.io/en/latest/markdown/podman.1.html#environment-variables
> 
> thanks
> 
> I think you should call /etc/qm/containers/containers.conf.d/qm_engine_container.conf"
This is what the environment does, once I'm placing this in the end of a config file at /etc/qm/containers/containers.conf.d/qm_image_tmp_dir.conf. I'm posting this at the bottom of the config file, right under the   "[engine]" section and its being addressed by the environment automatically.
image_copy_tmp_dir is the variable that should be changed from its default /var/tmp value to /var/tmp.dir so that the image could get pulled successfully inside the qm.

[engine]
image_copy_tmp_dir="/var/tmp.dir"

> 
> Please add to test README how to update qm container parameters
Will do.

tests/ffi/common/prepare.sh
nsednev: Added code as suggested.
Please review if needed.

tests/ffi/common/prepare.sh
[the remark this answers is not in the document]
nsednev: Would you like me to create a README as a file, within the folder or just a reference in the code like I did here?
https://github.com/nsednev/qm/blob/a94b15ede6ccd3104f565e9a52132f1d3f49bc0f/tests/ffi/common/prepare.sh#L62C3-L73C15

aesteve-rh:
@nsednev when you receive requests from reviewers, you should apply them over the commit that caused the issue.

Try following these steps:
1. Rebase your branch with an up-to-date local main. This will remove the merge commit that you've addded:
```
$ git rebase main
warning: skipped previously applied commit ab9b8c9
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
Successfully rebased and updated refs/heads/VROOM-22955.
```
Then you can do:
```
git rebase -i HEAD~4
```
You should see only your commits, keep the first pick, the other commits switch to squash:
image

Save and exit editing the file. Next you will be prompted with the commit message and body selection. Remove the additional text from the commits, and leave a meaningful header and body, and the SoB line.

Now you should be only one commit ahead of main. So you can just force push:
```
git push -f
```

Next time you get a review and want to change something in the last commit you can just do `git add` and then `git commit --amend --no-edit`. Or a rebase to edit a previous commit if the reviewer wants you to change a commit that is not the HEAD of your branch.

Let me know if you have any problems during the process.

aesteve-rh:
Well, almost... You've rebased over the last commit from main.

It's fine, happens. You'll probably need to do a soft reset of your commit, rebase from main, and commit your changes again. I can be of assist if you need. Ping me if so.

tests/ffi/common/prepare.sh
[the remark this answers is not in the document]
nsednev: Added README as requested.

tests/ffi/common/README.md
```diff
@@ -0,0 +1,21 @@
+Title: OStree images using podman pull to pull containers, larger than 1G.
```
Yarboa, line 1: I suggest to move this readme to https://github.com/containers/qm/tree/main/docs
You can create directory here like qm containers.conf with README


please look simple explanation, explain how to change the variables  in qm containers.conf
And a link to the section in tests for sample code.

nsednev: > I suggest to move this readme to https://github.com/containers/qm/tree/main/docs You can create directory here like qm containers.conf with README
> 

Done.

> please look simple explanation, explain how to change the variables in qm containers.conf And a link to the section in tests for sample code.

I've made changes to explanations in README, if I've got you right.

dougsland:
I will let the others review but to my eyes looks good, thanks for the efforts to make it great.

dougsland:
@aesteve-rh @Yarboa feel free to merge.

aesteve-rh:
lgtm, thanks!

aesteve-rh (approved):

PR #585, opened 2024-10-01
qm: introduction to qm dropin sub-packages(s)

dougsland:
QM developers are always looking for ways to help new users and engineers onboard more easily. With this in mind, they decided to break complex configurations into drop-in sub-package(s) when
 possible for adjusting the Podman engine configurations specifically
within the QM environment. Why? This approach provides an easy way to extend or override settings without modifying the core (original) configuration files.


dougsland:
Here output from the patch:
```
qm> git checkout remotes/origin/qm_dropin_img_tempdir -b qmdropin
qm> sed -i 's/Version: 0/Version: 1.0/g' rpm/qm.spec
qm> rm -rf rpmbuild && VERSION=1.0 make rpm
snip...
Wrote: /home/douglas/dropin/qm/rpmbuild/RPMS/noarch/qm-dropin-img-tempdir-1.0-1.fc40.noarch.rpm
Wrote: /home/douglas/dropin/qm/rpmbuild/RPMS/noarch/qm-1.0-1.fc40.noarch.rpm
```

inspecting:
```
qm> cd rpmbuild/RPMS/noarch/
qm> rpm2cpio qm-dropin-img-tempdir-1.0-1.fc40.noarch.rpm | cpio -div
./etc/qm/containers/containers.conf.d/qm_dropin_img_tempdir.conf
2 blocks

qm> $ cat etc/qm/containers/containers.conf.d/qm_dropin_img_tempdir.conf
# This configuration file was generated by qm-dropin sub-package:
#     - qm-dropin-img-tempdir
#
# It configures the temporary directory used by the container engine
# for image copy operations. This setting is useful for performance
# optimizations or managing disk space.
#
# How it works?
#=================
# [engine] <- section
# image_copy_tmp_dir <- Specifies the directory for temporary image storage.
[engine]
image_copy_tmp_dir="/var/tmp.dir"
``` 

To avoid the sub-package generation, add the following patch into Makefile:
```
diff --git a/Makefile b/Makefile
index 8e61e9b..283d434 100644
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ rpm: clean dist
        mkdir -p ${RPM_TOPDIR}/{RPMS,SRPMS,BUILD,SOURCES}
        cp ./rpm/v${VERSION}.tar.gz ${RPM_TOPDIR}/SOURCES
        rpmbuild -ba \
+               --without qm_dropin_img_tempdir \
                --define="_topdir ${RPM_TOPDIR}" \
                --define="version ${VERSION}" \
                ${SPECFILE}
```

Test generating the same rpm, now only generate one rpm and src rpm:
```
rm -rf rpmbuild/ && VERSION=1.0 make rpm
SNIP...
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/douglas/dropin/qm/rpmbuild/BUILDROOT/qm-1.0-1.fc40.x86_64
Wrote: /home/douglas/dropin/qm/rpmbuild/SRPMS/qm-1.0-1.fc40.src.rpm
Wrote: /home/douglas/dropin/qm/rpmbuild/RPMS/noarch/qm-1.0-1.fc40.noarch.rpm
Executing(%clean): /bin/sh -e /var/tmp/rpm-tmp.HMNZ0U
```

dougsland:
`You may wondering...  Douglas, it's just one file, why a sub-package?`
In this example, it's just one file but drop-in support multiple files, so users (or us) we can have a sub-package with 10-qm-blah.conf 20-qm-bleh.conf etc.  More info here: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html

dougsland:
@nsednev @pengshanyu is it something are you aware this error? Everything I don't want is bring a new issue to us.

dougsland:
```
                if test -d /var/qm -a -d /etc/qm ; then
                        info_message "QM Enabled and not Active"
                        info_message "=============================="
                        exit 1
                fi
```
maybe remove this exit 1 ^ from: `tests/e2e/set-ffi-env-e2e`

dougsland:
> ```
>                 if test -d /var/qm -a -d /etc/qm ; then
>                         info_message "QM Enabled and not Active"
>                         info_message "=============================="
>                         exit 1
>                 fi
> ```
> 
> maybe remove this exit 1 ^ from: `tests/e2e/set-ffi-env-e2e`

Looks like not working (seems stuck in a loop in CI/CD), removing this patch.

dougsland:
> > ```
> >                 if test -d /var/qm -a -d /etc/qm ; then
> >                         info_message "QM Enabled and not Active"
> >                         info_message "=============================="
> >                         exit 1
> >                 fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > maybe remove this exit 1 ^ from: `tests/e2e/set-ffi-env-e2e`
> 
> Looks like not working (seems stuck in a loop in CI/CD), removing this patch.

Ok, not even related to the previous patch. CI/CD seems weird. :(

dougsland:
Ok, the CI/CD URL is not even found, I got 404: https://artifacts.dev.testing-farm.io/4a314a63-f306-4201-96f5-c6f1859a2877

dougsland:
I don't think it's related to this patch but let's wait another maint. review.

Yarboa (commented):
MHO, this should be documented and not related to rpm
Since drop-in files are used for Customizing the base configuration

dougsland:
> MHO, this should be documented and not related to rpm Since drop-in files are used for Customizing the base
>  configuration

Let me try to improve my explanation:

Packaging the drop-in files within the RPM ensures that users don't need to manually create or configure these files. This can improve the user experience, particularly for less experienced users if the "action" is repetitive, example "changing the tmp dir for engine" or "changing a configuration for qm window manager" or "a vendor/suppliers shipping their specific podman engine configuration".

With that in mind, keeping drop-in files out of the RPM can reduce the complexity of maintaining the package, especially if the configurations are expected to change frequently (which I believe it's not the case here, as we want to have "flavors" of qm sub-packages for podman engine configuration so users can just dnf/rpm install and voi-la ready to use).

@Yarboa it sounds better now? What do you think ?

dougsland:
In the example, I attached, we could do:

```
   is_ostree:
        dnf install qm-dropin-img_tempdir -y
```

or if the vendor/user would like to try later they can do it without worry and later remove it.

Yarboa:
> In the example, I attached, we could do:
> 
> ```
>    is_ostree:
>         dnf install qm-dropin-img_tempdir -y
> ```
> 
> or if the vendor/user would like to try later they can do it without worry and later remove it. At least, that's the idea.

I choose not to die on that one, lets hear @aesteve-rh or others

dougsland:
> > In the example, I attached, we could do:
> > ```
> >    is_ostree:
> >         dnf install qm-dropin-img_tempdir -y
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or if the vendor/user would like to try later they can do it without worry and later remove it. At least, that's the idea.
> 
> I choose not to die on that one, lets hear @aesteve-rh or others

Funny enough, I actually learned that phrase from you @Yarboa "I choose not to die on that one" ^^^ !!! 👍  My last argument, enterprise vendors prefer changes that have rpm signed. All yours @aesteve-rh maybe @sandrobonazzola  too if he is around.

docs/devel/README.md
```diff
@@ -244,3 +246,20 @@ bash-5.1# podman --root /usr/share/containers/storage pull alpine
...
+
+QM developers are always looking for ways to help new users and engineers onboard more easily. With this in mind, they decided to break complex configurations into drop-in sub-package(s) when possible for adjusting the Podman engine configurations specifically within the QM environment. Why? This approach provides an easy way to extend or override settings without modifying the core (original) configuration files.
+
+### Creating your own drop-in QM sub-package
+
+We recommend using the existing drop-in files as a guide and adapting them to your specific needs. However, here are the step-by-step instructions:
+
+1) Create a drop-in file in the directory: `etc/qm/containers/containers.conf.d`
+2) Add it as a sub-package to `rpm/qm.spec`
+3) Test it by running: `rm -rf rpmbuild/ && VERSION=YOURVERSIONHERE make rpm`
```
aesteve-rh, line 260: Same as above:
```suggestion
3) Test it by running: `make clean && VERSION=YOURVERSIONHERE make rpm`
```

aesteve-rh:
In general, this being a feature for developers only, I feel it's fine. I am not against this patch.

I'm not convinced that generating the subpackage should be the default behavior. And I guess we should think of a way to tweak the makefile to avoid having to edit it manually to chose whether you want the subpackage or not.

docs/devel/README.md
```diff
@@ -46,7 +48,7 @@ Make necessary changes using your preferred editor (e.g., vi/emacs).
 **4.** Build the RPM
 
 ```bash
-VERSION=1.0 make rpm
+make clean && VERSION=1.0 make rpm
```
aesteve-rh, line 51: Actually, looking at the makefile target, `rpm` depends on `clean`. Why did you decide to add the `rm -rf` before? It should've been cleaned either way before `rpmbuild`.

dougsland: it was not cleaning the old file. I think it's save to keep it in the docs.

aesteve-rh (approved):
lgtm

PR #591, opened 2024-10-03
[WIP] qm subpackage: Enabling audio in QM container and nested containers with mount bind /dev/snd

dougsland:
Enabling audio via mount bind /dev/snd into QM container and nested container.


dougsland:
I just need to fully test this one before merge but the concept is that. Should work.

dougsland:
Still testing the solution, moving the title to WIP.

dougsland:
Sent a new PR, deprecated this one.

PR #593, opened 2024-10-03
Add Fedora 41 and drop Fedora 38 from Packit config

nforro:
Is there a reason why you don't want to use aliases such as `fedora-all`, `fedora-branched` or `epel-all`?

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -112,9 +112,9 @@ jobs:
     update_release: false
     dist_git_branches:
       - fedora-rawhide
+      - fedora-41
       - fedora-40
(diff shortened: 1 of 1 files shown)


dougsland:
> Is there a reason why you don't want to use aliases such as `fedora-all`, `fedora-branched` or `epel-all`?

No reason, please open a bug and if you are interested in work in this one, let's assign to you :)

nforro:
> No reason, please open a bug and if you are interested in work in this one, let's assign to you :)

I've opened https://github.com/containers/qm/pull/621.

PR #604, opened 2024-10-06
Added fix to c9s

Yarboa:
Under CI special case exist.
rpm qm is installed in c9s machine, but setup
script that emulate the QM environment is not set


dougsland:
thanks a lot @Yarboa

PR #606, opened 2024-10-07
qm sound sub-package: enable audio

dougsland:
Enable audio in QM container and nested containers


dougsland:
Later, we will have one sub-package with kvm and other with kvm + virtio enabling. The wayland is almost there I just need to spend time on it. Users will be able to enable/disable feature by dnf/rpm with signed repos or using official distros.

Makefile
```diff
@@ -85,14 +86,20 @@ qm_dropin_mount_bind_ttyUSB0: ##     - Creates a QM RPM sub-package to mount bin
 	tools/version-update -v ${VERSION}
 	$(MAKE) VERSION=${VERSION} rpm
 
+.PHONY: qm_dropin_mount_bind_sound
+qm_dropin_mount_bind_sound: ##       - Creates a QM RPM sub-package to mount bind /dev/snd in the nested containers
+	sed -i 's/%define enable_qm_mount_bind_sound 0/%define enable_qm_mount_bind_sound 1/' ${SPECFILE}
+	tools/version-update -v ${VERSION}
```
aesteve-rh, line 92: If the intention is to allow `make` to update the version, we could move it to the `rpm` target instead. So that it does not need to be explicitly added for each other (subpackage) target.

dougsland: @aesteve-rh agreed. let me know if you would like to send a patch.

aesteve-rh: Can do.

PR #608, opened 2024-10-08
qm sub-package Kernel-based Virtual Machine (KVM)

dougsland:
The QM sub-package KVM includes drop-in configuration that enables the integration of Kernel-based Virtual Machine (KVM) management into the QM (Quality Management) container environment. This configuration allows users to easily configure and manage KVM virtual machines within the QM system, streamlining virtualization tasks in containerized setups.

Use: make qm_dropin_mount_bind_kvm

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
     - [Installing QM sub-packages](Installing-qm-sub-packages)
     - [Removing QM sub-packages](Removing-qm-sub-packages)
     - [Creating your own drop-in QM sub-package](Creating-your-own-drop-in-QM-sub-package)
+    - [QM sub-package kvm](QM-sub-package-kvm)
     - [QM sub-package Sound](QM-sub-package-Sound)
 3. [SELinux Policy](#selinux-policy)
 4. [BlueChi](#bluechi)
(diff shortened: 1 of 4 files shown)


aesteve-rh (approved):

README.md
```diff
@@ -288,6 +289,134 @@ Example changing the spec and triggering the build via make (feel free to automa
...
+$host> sudo dnf install rpmbuild/RPMS/noarch/qm_mount_bind_kvm-0.6.7-1.fc40.noarch.rpm
+$host> sudo podman restart qm  # if you have qm already running
+$host> sudo dnf --installroot /usr/lib/qm/rootfs/ install virt-install libvirt-daemon libvirt-daemon-qemu libvirt-daemon-kvm -y
+
+# Copy default network settings to /root dir inside QM (/usr/lib/qm/rootfs/root)
+$host> sudo cp /usr/share/libvirt/networks/default.xml /usr/lib/qm/rootfs/root
+
+Step 2: Preparing cloudinit files inside QM (/usr/lib/qm/rootfs/root)
+
+# Cloud-init files
```
Yarboa, line 312: Why are you using cloud-init? i assume it is for the guest? 
Is it only for the password?  

I prefer to use this command after that one 
```
virt-customize --uninstall cloud-init --root-password password:Aa1234 -a ./<fedora-image>.qcow2
```

https://github.com/containers/qm/pull/608/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R321

README.md
```diff
@@ -288,6 +289,134 @@ Example changing the spec and triggering the build via make (feel free to automa
...
+$ wget -O /usr/lib/qm/rootfs/root/Fedora-Cloud-Base-Generic.qcow2 https://download.fedoraproject.org/pub/fedora/linux/releases/40/Cloud/x86_64/images/Fedora-Cloud-Base-Generic.x86_64-40-1.14.qcow2
+
+# Generate the cloud-init.iso and move it to /usr/lib/qm/rootfs/var/lib/libvirt/images/
+$host> cloud-localds cloud-init.iso user-data meta-data
+$host> mv cloud-init.iso /usr/lib/qm/rootfs/var/lib/libvirt/images/
+
+# Change permission to qemu:qemu
+$host> chown qemu:qemu /usr/lib/qm/rootfs/var/lib/libvirt/*
+
+Step 3: Starting libvirtd and checking if it's active inside QM
```
Yarboa, line 337: did you hit selinux issues?

README.md
```diff
@@ -288,6 +289,134 @@ Example changing the spec and triggering the build via make (feel free to automa
...
+Below example step by step:
+
+Step 1:  clone QM repo, install libvirt packages, prepare some files inside QM and start the libvirt daemon.
+
+```bash
+$host> git clone https://github.com/containers/qm.git && cd qm
+$host> make qm_dropin_mount_bind_kvm
+$host> sudo dnf install rpmbuild/RPMS/noarch/qm_mount_bind_kvm-0.6.7-1.fc40.noarch.rpm
+$host> sudo podman restart qm  # if you have qm already running
+$host> sudo dnf --installroot /usr/lib/qm/rootfs/ install virt-install libvirt-daemon libvirt-daemon-qemu libvirt-daemon-kvm -y
```
Yarboa, line 305: prefer to use this one 
```
sudo dnf install qemu
```
qemu-system-x86_64 --version

aesteve-rh: Technically, `qemu` should be containerized. This is nice for testing that the kvm device has been mounted correctly, but it is not the way to run a VM inside the QM.

Yarboa: Thanks @aesteve-rh ah, yes, the whole things should be containerized,
Step 1:
 is running qemu in qm
 Step 2:
 qemu containerized in qm    
Is it? 

One more, 
Do you think we should use virt-* packages?

aesteve-rh: Step 2 should be supported already, so we can aim for that directly.

> Do you think we should use virt-* packages?

You mean for this documented example or in general? Probably, as a response to both, it should not be necessary to use them. I agree with you in that qemu would be preferred.

README.md
dougsland: yes, must be addressed in a next patch.

README.md
dougsland: @aesteve-rh yes, should work like that. @aesteve-rh can you share more about your vision so all of us are in the same page?

README.md
dougsland: sounds a good plan.  If we merge can you test and improve it for us?

dougsland:
> One more,
> Do you think we should use virt-* packages?

Yes.

dougsland:
> Thanks @aesteve-rh ah, yes, the whole things should be containerized,

Ah @aesteve-rh you meant ContinerFile? Yes and shipped inside the RPM @Yarboa  (in a next patch). Let's merge what we have and keep improving.

README.md
aesteve-rh: I am only saying that workloads inside the qm are supposed to be containerized. Like, the image should be stored inside the qm and launched through a quadlet file. That allows to set rules to ensure FFI, control the system's surface that the process has access to, and also allows applying SELinux labels to allow only a correct subset of operations.

For developers, it may suffice to install all those packages directly on the qm container and test it there. But imo, this is not how virtualization should be documented for users.

Yarboa (approved):
Extra work based on the commens here #609

README.md
dougsland: @aesteve-rh completed agreed. It's just a documentation.

PR #612, opened 2024-10-10
Adding VARIABLES for subpackages build

Yarboa:
Adding support in build env vars 

This set rpmbuild --define vars in Makefile. 
remove define vats for building subpackages

Toggle package to build by define export variable=1


dougsland:
The patch looks good, but CI/CD is complaining, need investigation.

PR #613, opened 2024-10-11
add support for tmt prune

pengshanyu:
When using tmt `prune` (Copy only immediate directories of executed tests and their required files. Added in version 1.29.) in the test plan, need to specify the location of the included file, otherwise an error will occur.

### DIFF
--- a/tests/ffi/main.fmf
+++ b/tests/ffi/main.fmf
@@ -1,0 +1,8 @@
+/:
+    inherit: false
+
+require:
+  - type: file
+    pattern:
+      - /tests/e2e/lib/utils
+      - /tests/ffi/common/prepare.sh


tests/ffi/main.fmf
```diff
@@ -0,0 +1,8 @@
+/:
```
dougsland, line 1: `/: ` is this required?

dougsland: In general looks good

Yarboa (approved):
LGTM

tests/ffi/main.fmf
pengshanyu: Yes, /: with inherit: false, it is the fmf syntax.

PR #620, opened 2024-10-15
github-actions: add test to build dropin targets

dougsland:
Just test if make dropin targets works


dougsland:
I am sure soon we will do better validation than that but for now, is good to have something to guarantee we don't break the subpackages.

Yarboa:
> I am sure soon we will do better validation than that but for now, is good to have something to guarantee we don't break the subpackages.

Please explain what is the goal of this PR

dougsland:
> > I am sure soon we will do better validation than that but for now, is good to have something to guarantee we don't break the subpackages.
> 
> Please explain what is the goal of this PR

Create a github actions to execute a loop to all existing subpackages a

> > I am sure soon we will do better validation than that but for now, is good to have something to guarantee we don't break the subpackages.
> 
> Please explain what is the goal of this PR

Create a github actions that execute `make qm_dropin_<NAME_HERE>`. Please note `<NAME_HERE>` is dynamic the number of subpackages can increase or decrease with time.

dougsland:
I should be back here soon, just finishing some work in the subpackage of Wayland. If anyone would like to continue this work, feel free.

dougsland:
Resolves: https://github.com/containers/qm/issues/628

PR #621, opened 2024-10-16
Fix Packit configuration

nforro:


.packit.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     # x86_64 is assumed by default
     # qm is noarch so we only need to test on one arch
     targets:
-      fedora-rawhide: {}
+      fedora-development: {}
```
nforro, line 21: Configuring only `fedora-rawhide` and `fedora-latest` means the job will run for `rawhide` and `f40`, but not `f41`. But if that's what you want, I can revert this change.

.packit.yaml
```diff
@@ -37,15 +37,9 @@ jobs:
...
-      fedora-40-x86_64: {}
-      fedora-39-aarch64: {}
-      fedora-39-ppc64le: {}
-      fedora-39-x86_64: {}
-      fedora-38-aarch64: {}
-      fedora-38-ppc64le: {}
-      fedora-38-x86_64: {}
+      fedora-stable-aarch64: {}
+      fedora-stable-ppc64le: {}
+      fedora-stable-x86_64: {}
```
nforro, line 42: I assume the architectures are listed here to avoid building on `s390x`, right?

.packit.yaml
```diff
@@ -109,32 +103,25 @@ jobs:
 
   - job: propose_downstream
     trigger: release
-    update_release: false
```
nforro, line 112: This option doesn't apply to `propose_downstream` and has no effect.

.packit.yaml
```diff
@@ -109,32 +103,25 @@ jobs:
...
     dist_git_branches:
-      - fedora-all-aarch64
-      - fedora-all-ppc64le
-      - fedora-all-x86_64
-      - epel-all-aarch64
-      - epel-all-ppc64le
-      - epel-all-x86_64
```
nforro, line 119: These are dist-git branch names, not build targets, there is no architecture.

.packit.yaml
```diff
@@ -109,32 +103,19 @@ jobs:
...
     trigger: commit
     dist_git_branches:
       # rawhide updates are created automatically
-      - fedora-all-aarch64
-      - fedora-all-ppc64le
-      - fedora-all-x86_64
-      - epel-all-aarch64
-      - epel-all-ppc64le
-      - epel-all-x86_64
+      - fedora-branched
```
nforro, line 120: `fedora-branched` expands to all Fedora versions except `rawhide`.

.packit.yaml
```diff
@@ -109,32 +103,19 @@ jobs:
...
-    update_release: false
     dist_git_branches:
-      - fedora-all-aarch64
-      - fedora-all-ppc64le
-      - fedora-all-x86_64
-      - epel-all-aarch64
-      - epel-all-ppc64le
-      - epel-all-x86_64
+      - fedora-all
+      - epel-9
```
nforro, line 108: I'm only listing `epel-9` here because there is only `epel9` branch in dist-git. Also, `epel-all` currently expands to `epel-8`, `epel-9` and `epel-10.0`, which is [problematic](https://github.com/packit/packit/issues/2438), so if you want to do releases for `epel10`, it is necessary (after requesting the branch in dist-git) to list it here explicitly as `epel-10` or `epel10`.

.packit.yaml
```diff
@@ -37,15 +37,9 @@ jobs:
...
-      fedora-39-x86_64: {}
-      fedora-38-aarch64: {}
-      fedora-38-ppc64le: {}
-      fedora-38-x86_64: {}
+      fedora-stable-aarch64: {}
+      fedora-stable-ppc64le: {}
+      fedora-stable-x86_64: {}
       centos-stream-9-aarch64:
         additional_repos:
           - https://dl.fedoraproject.org/pub/epel/9/Everything/aarch64/
```
nforro, line 45: Are you doing this to build explicitly on CentOS Stream 9 and not RHEL 9? Otherwise you could simply use `epel9-aarch64` with no `additional_repos`.

.packit.yaml
dougsland: yes

.packit.yaml
dougsland: @nforro perfect. Would mind to send us another contribution? :)

.packit.yaml
dougsland: No, that's fine.

dougsland:
Thanks for the fixed @nforro , my mistake. I thought we should list one bye one instead use branched string.

.packit.yaml
nforro: Will do.

PR #626, opened 2024-10-18
qm wm: add more mount bind for auto active login

dougsland:
loginctl was not able to auto autoactive and it
was showing Permission Denied when executing
loginctl acticate <number session>. This
patch mount bind the required path to qm.

### DIFF
--- a/qm-windowmanager/etc/containers/systemd/session-activate.container
+++ b/qm-windowmanager/etc/containers/systemd/session-activate.container
@@ -8,6 +8,7 @@ Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/dbus/system_bus_socket
 Image=quay.io/qm-images/wm_active_session:latest
 SecurityLabelType=qm_container_wayland_t
 Volume=/run/systemd:/run/systemd:ro
+Volume=/etc/dbus-1/system.d:/etc/dbus-1/system.d:ro
 Volume=/run/dbus/system_bus_socket:/run/dbus/system_bus_socket
 Volume=/run/user/0:/run/user/0
 


aesteve-rh:
Which error did you get if the volume is missing? I did not need it in my tests, so let's see what's different.

dougsland:
> Which error did you get if the volume is missing? I did not need it in my tests, so let's see what's different.

Permission Denied. In AutoSD Demo I assume it's based on CentOS and not Fedora? Also based on KVM here is pure podman so a lot of tests now for us to start to covering.

dougsland:
I think autosd demo also have good stuff to cover in this dir to include this rule of auto mount there:
```
/etc/dbus-1/system.d# ls
avahi-dbus.conf                          oddjob.conf                                      org.freedesktop.problems.daemon.conf
blivet.conf                              oddjob-mkhomedir.conf                            org.freedesktop.realmd.conf
com.redhat.lvmdbus1.conf                 org.fedorahosted.certmonger.conf                 org.freedesktop.RealtimeKit1.conf
com.redhat.NewPrinterNotification.conf   org.fedoraproject.Setroubleshootd.conf           org.gnome.GConf.Defaults.conf
com.redhat.PrinterDriversInstaller.conf  org.fedoraproject.SetroubleshootFixit.conf       org.selinux.conf
cups.conf                                org.fedoraproject.SetroubleshootPrivileged.conf  pulseaudio-system.conf
dbus-abrt.conf                           org.freedesktop.Flatpak.SystemHelper.conf        teamd.conf
dnsmasq.conf                             org.freedesktop.GeoClue2.Agent.conf              vss-dbus.conf
gdm.conf                                 org.freedesktop.GeoClue2.conf                    vss-dbus.conf.bak
nm-ssh-service.conf                      org.freedesktop.ModemManager1.conf               wpa_supplicant.conf
```

dougsland:
However, I am still testing to make sure it was not related to`--privileged` tested and somehow buffered in the system. 
/hold

dougsland:
Also, looks like the SELinux policies in Fedora seems different from the CentOS ones, we need to have a specific issue to research and update the rules.

dougsland:
I don't believe it's safe to merge, closing for now.

PR #633, opened 2024-10-27
Add ROS2 subpackage into QM

dougsland:
Just initial work to support ROS2 - "The Robot Operational System" (The middleware for robots) out of box into QM.

ContainerFile by Leonardo Rossetti <lrossett@redhat.com> with few mods by me.

See-Also: https://ros.org/


dougsland:
it seems the diff scanner is not working well, it's getting stuck, not first time today.

PR #643, opened 2024-10-30
packit: add fedora-latest

dougsland:
Lets add fedora to the tests


Yarboa:
@dougsland Manual spike work should be done for fedora testing.
I think it worth opening an issue for it

Issue #629, opened 2024-10-25
improve agent-flood


pengshanyu:
The following problem sometimes occur:
```
$ podman exec qm systemctl status bluechi-tester-1

× bluechi-tester-1.service - bluechi-tester-1
     Loaded: loaded (/etc/containers/systemd/bluechi-tester-1.container; generated)
     Active: failed (Result: timeout) since Wed 2024-10-23 07:15:19 UTC; 18s ago
    Process: 151 ExecStart=/usr/bin/podman run --name=systemd-bluechi-tester-1 --cidfile=/run/bluechi-tester-1.cid --replace --rm --cgroups=split --network=host --sdnotify=conmon -d dir:/var/lib/containers/registry/tools-ffi:latest /root/tests/FFI/bin/bluechi-tester --url=tcp:host=10.26.28.155,port=842 --nodename=bluechi-tester-1 --numbersignals=11111111 --signal=JobDone (code=exited, status=0/SUCCESS)
    Process: 176 ExecStopPost=/usr/bin/podman rm -v -f -i --cidfile=/run/bluechi-tester-1.cid (code=exited, status=0/SUCCESS)
   Main PID: 151 (code=exited, status=0/SUCCESS)
        CPU: 2min 4.318s

Oct 23 07:13:24 27c251633eae systemd[1]: Starting bluechi-tester-1...
Oct 23 07:13:24 27c251633eae bluechi-tester-1[151]: Getting image source signatures
Oct 23 07:13:24 27c251633eae bluechi-tester-1[151]: Copying blob sha256:191c327d78ba03afe30fb233a6b18249870076d0ddee6c7801206fb064908857
Oct 23 07:13:24 27c251633eae bluechi-tester-1[151]: Copying blob sha256:d65c66ea6e0ce9c41b9b9119ab07b00bb97f4de33bb0ce95558991ebb961af83
Oct 23 07:14:54 27c251633eae systemd[1]: bluechi-tester-1.service: start operation timed out. Terminating.
Oct 23 07:15:19 27c251633eae systemd[1]: bluechi-tester-1.service: Failed with result 'timeout'.
Oct 23 07:15:19 27c251633eae systemd[1]: Failed to start bluechi-tester-1.
Oct 23 07:15:19 27c251633eae systemd[1]: bluechi-tester-1.service: Consumed 2min 4.318s CPU time.
```
[...]

PR #646, opened 2024-10-31
change bluechi controller port to 8420

pengshanyu:
842 is a privileged port of bluechi, so change it to  8420
Solved:  https://github.com/containers/qm/issues/629

### DIFF
--- a/tests/e2e/set-ffi-env-e2e
+++ b/tests/e2e/set-ffi-env-e2e
@@ -190,7 +190,7 @@ setup_qm_services() {
   cat > /etc/bluechi/controller.conf << 'EOF'
 [bluechi-controller]
 AllowedNodeNames=qm.localrootfs,localrootfs
-ControllerPort=842
+ControllerPort=8420
 LogLevel=INFO
 LogTarget=journald
 LogIsQuiet=false
@@ -198,6 +198,7 @@ EOF
 cat > /etc/bluechi/agent.conf.d/00-default.conf << 'EOF'
 [bluechi-agent]
 NodeName=localrootfs
+ControllerPort=8420
 EOF
 
 controller_host_ip=$(hostname -I | awk '{print $1}')
--- a/tests/e2e/set-ffi-env-e2e
+++ b/tests/e2e/set-ffi-env-e2e
@@ -201,11 +201,13 @@ NodeName=localrootfs
 ControllerPort=8420
 EOF
 
+# Insert ControllerHost and ControllerPort into /etc/qm/bluechi/agent.conf.d/agent.conf
 controller_host_ip=$(hostname -I | awk '{print $1}')
 qm_bluechi_agent_config_file="/etc/qm/bluechi/agent.conf.d/agent.conf"
 if [[ -f "${qm_bluechi_agent_config_file}" ]]; then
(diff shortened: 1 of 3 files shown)


pengshanyu:
qm-tier-0 test failed because `qm.localrootfs` is not connected
To solve this issue, we need to insert `ControllerPort=8420` to /etc/qm/bluechi/agent.conf.d/agent.conf
But I noticed that /etc/qm/bluechi/agent.conf.d/agent.conf is generated in the `bluechiSetup()` function in setup

Do you think it is better to modify setup or insert `ControllerPort=8420` into the /etc/qm/bluechi/agent.conf.d/agent.conf  after agent.conf is generated?
Hi @dougsland @Yarboa, could you help to take a look? Thanks.

Yarboa:
I do not understand the issue,
https://github.com/eclipse-bluechi/bluechi/blob/main/doc/man/bluechi-controller.conf.5.md
Port was not changed, so why there is a need to change it? 

And with your change ci is failing on connectivity
https://artifacts.dev.testing-farm.io/ec019034-09be-4389-aaad-a8951424bb7e/

Can you please explain the cause of failure? where ids it happen and why?

pengshanyu:
Bluechi developers suggest that it would be best to use some port >1024, e.g. 8420, because 842 is a privileged port. 
The failure is because I changed the port to 8420 in ffi-tools, so we need to add `ControllerPort=8420` into /etc/qm/bluechi/agent.conf.d/agent.conf accordingly

Yarboa:
> Bluechi developers suggest that it would be better to use some port >1024, e.g. 8420, because 842 is a privileged port. The CI failure is because I changed the port to 8420 in ffi-tools, so we need to add `ControllerPort=8420` into /etc/qm/bluechi/agent.conf.d/agent.conf accordingly

Please refer this from here  
[root@localhost ~]# cat /etc/bluechi/controller.conf
[bluechi-controller]
AllowedNodeNames=host,qm.host

https://autosd.sig.centos.org/AutoSD-9/nightly/sample-images/
use qa image, 

I would change it here, after port is changed in image manifest
To avoid deviations

@dougsland you thoughts are welcome

tests/e2e/set-ffi-env-e2e
```diff
@@ -190,21 +190,24 @@ setup_qm_services() {
...
+ControllerPort=8420
 EOF
 
+# Insert ControllerHost and ControllerPort into /etc/qm/bluechi/agent.conf.d/agent.conf
 controller_host_ip=$(hostname -I | awk '{print $1}')
 qm_bluechi_agent_config_file="/etc/qm/bluechi/agent.conf.d/agent.conf"
 if [[ -f "${qm_bluechi_agent_config_file}" ]]; then
     if ! grep "ControllerHost=${controller_host_ip}" "${qm_bluechi_agent_config_file}" >/dev/null; then
         sed -i '$a \ControllerHost='"${controller_host_ip}"'' ${qm_bluechi_agent_config_file}
+        sed -i '$a \ControllerPort=8420' ${qm_bluechi_agent_config_file}
```
pengshanyu, line 210: Do you think it is better to move `ControllerPort=8420` to `bluechiSetup()` of `setup`

pengshanyu:
Once the port is changed in the qa image, we will support this change.

PR #647, opened 2024-10-31
packit: add fedora latest

dougsland:


Yarboa:
I would advise you the following:
- Check locally why there is not connectivity from the agent side inside QM
- Add special used case for fedora disk partitioning, there is a tests checking it

dougsland:
> I would advise you the following:
> 
> * Check locally why there is not connectivity from bluechi-agent side inside QM
> * Add special used case for fedora disk partitioning, there is a tests checking it

Hey Yariv,

I guess it's something else, please look the setup logs, it says no fedora repo available, we are working on that. After that one solve and the system is able to install the packages we can see what else is missing.

dougsland:
@nforro any idea?

dougsland:
@nforro let's continue the journey tomorrow? https://artifacts.dev.testing-farm.io/6020e5b7-4555-489a-beb4-cfe187d122df/work-tier-0wj55elim/plans/e2e/tier-0/execute/data/guest/default-0/tests/qm-sanity-test/check_qm_setup_succeeds-6/output.txt

PR #648, opened 2024-11-03
General improvements

dougsland:
- ros2-rolling: remove the restart action
- docs: debugging quadlets


dougsland:
skipping diff scan as it's readme and removing line.

PR #653, opened 2024-11-08
add NoRenesas tag to /tests/ffi/disk

pengshanyu:
Add a NoRenesas tag to /tests/ffi/disk to disable it when there is no separated /var disk partition for QM.


dougsland:
thanks @pengshanyu !

PR #655, opened 2024-11-09
setup: for fedora use --use-host-config for dnf repos

dougsland:
It will solve the recent CI/CD setup install complain about not able to find packages.


dougsland:
--use-host-config is a new feature only available in dnf Fedora.

PR #672, opened 2024-12-04
spec: remove the dirs creations

dougsland:
It should go directly in the subpackages which requires it. On top of that it BROKE autosd installer, the installer reports  "mkdir -EEXIST".

/usr/lib/qm/rootfs/etc
/usr/lib/qm/rootfs/etc/containers
/usr/lib/qm/rootfs/etc/containers/system


dougsland:
cc @Yarboa @alexlarsson first patch of the series. Going to work in the subpackages.

Yarboa (approved):
LGTM

PR #673, opened 2024-12-05
check /var partition before run /tests/ffi/disk

pengshanyu:
Added a function `check_var_partition` to check if a separate /var disk partition exists before executing the test, so as not to block the execution of other test cases after it.


pengshanyu:
Hi @Yarboa , could you please help to review this PR, thanks.

Yarboa (approved):
LGTM

PR #678, opened 2024-12-19
Adding comments for container options in qm.container

ilimarx:
Add comments to [container] section under qm quadlet file
upstream doc reference added to each quadlet param,

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -35,10 +35,28 @@ LimitNOFILE=65536
 TasksMax=50%
 
 [Container]
+# AddCapability
+# -------------
+# Add these capabilities, in addition to the default Podman capability set, to the container.
+# If set to all, grants all capabilities to the container, increasing flexibility but significantly
+# reducing security.
+# For details see: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#addcapability
 AddCapability=all
+
+# Unmask
+# -------
+# Specify the paths to unmask separated by a colon. unmask=ALL or /path/1:/path/2, or shell expanded paths (/proc/*):
+# If set to ALL, Podman will unmask all the paths that are masked or made read-only by default.
+# For details see: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#unmask
 Unmask=ALL
 SecurityLabelNested=true
 SeccompProfile=/usr/share/qm/seccomp.json
+
+# PidsLimit
+# ---------
+# Disables the PID limit for the container by setting it to -1.
+# Without a limit, the container can spawn unlimited processes, potentially exhausting system resources.
+# For details see: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#pidslimit
 PidsLimit=-1
 
 # Comment DropCapability this will allow FFI Tools to surpass their defaults.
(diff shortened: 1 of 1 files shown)


qm.container
```diff
@@ -35,10 +35,24 @@ LimitNOFILE=65536
 TasksMax=50%
 
 [Container]
+# AddCapability
```
Yarboa, line 38: I would recommend here to use man quadlet also for details

qm.container
```diff
@@ -35,10 +35,24 @@ LimitNOFILE=65536
...
+# -------------
+# Grants all capabilities to the container, increasing flexibility but significantly
+# reducing security.
 AddCapability=all
```
Yarboa, line 42: same as above

qm.container
```diff
@@ -35,10 +35,24 @@ LimitNOFILE=65536
...
+
+# Unmask
+# -------
+# Unmasks all systemd services for the container, overriding masking that prevents
+# access to specific services.
```
Yarboa, line 47: same as above

qm.container
```diff
@@ -49,7 +63,16 @@ AddDevice=-/dev/fuse
...
 Network=private
+
+# ReadOnly
+# --------
+# Makes the container's filesystem read-only, enhancing security by preventing modifications.
 ReadOnly=true
+
+# Rootfs
+# ------
+# Defines the root filesystem location for the container.
```
Yarboa, line 74: I would specify the QM rootfs partition

qm.container
```diff
@@ -49,7 +63,16 @@ AddDevice=-/dev/fuse
...
+# ReadOnly
+# --------
+# Makes the container's filesystem read-only, enhancing security by preventing modifications.
 ReadOnly=true
+
+# Rootfs
+# ------
+# Defines the root filesystem location for the container. 
+# The '${ROOTFS}' variable should point to a valid filesystem path.
 Rootfs=${ROOTFS}
```
Yarboa, line 76: see previous one is it a container or partitioned filesystem?

qm.container
Yarboa: see lower explanation related the usage of container

qm.container
Yarboa: same as below

qm.container
Yarboa: Consider adding this link to upstream doc
https://docs.podman.io/en/v4.4/markdown/podman-systemd.unit.5.html#addcapability

ilimarx: Added a link to the comment for AddCapability option. Not sure about the "upstream doc", I was told to write comments for options.

qm.container
ilimarx: Changed to partition

qm.container
Yarboa: Thanks, last comments related to that,

Consider thinking of man command , repo is upstream code, that could run against different podman versions, this way, in case of qm failures due to quadlet file, user can check the installed qudlet manual and refer the supported capabilities.

qm.container
Yarboa: If some on told you, this is the place to comment it

Yarboa:
@ilimarx please squash all commits. this change should be one commit
There are 6 commits for one change.
do not forget to verify b4 force push that in your git log your author is filled
and run the precommit checks commands before pushing it again

qm.container
pbrilla-rh: Instead of 4.4 link, if required would be much better to link as: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#addcapability

qm.container
ilimarx: I considered to write `man quadlet | awk '/AddCapability=$/,/^$/'` as you proposed, but the problem is that it cuts the description up to an empty line. Unfortunately, there are empty lines just in the middle of the AddCapability description. I tried to make an awk that will work better, but the command will be kind of long and scary, so I decided to leave links to the upstream doc (but the latest, as @pbrilla-rh wrote above)

qm.container
ilimarx: Should look nicer now. Also added links to other options.

Yarboa (approved):
LGTM,
It seems there is an issue with TestingFarm. related to `test --identifier e2e-ffi`
not related to PR

PR #679, opened 2024-12-28
[WIP] add libkrun support

dougsland:
initial code to add libkrun support

### DIFF
--- a/tools/qm-change-podman-runtime
+++ b/tools/qm-change-podman-runtime
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# Script to dynamically change Podman runtime
+# Usage: /usr/share/qm/qm-change-podman-runtime <runtime-name> <runtime-binary-paths>
+
+set -e
+
+# Usage function
+usage() {
+  echo "Usage: /usr/share/qm/qm-change-podman-runtime <runtime-name> <runtime-binary-paths>"
+  echo
+  echo "Examples:"
+  echo "  /usr/share/qm/qm-change-podman-runtime krun /usr/bin/krun"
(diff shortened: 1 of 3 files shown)


dougsland:
Hi @giuseppe @slp @rhatdan  after running the tool, I still see crun as OCIRuntime via podman info, Am I missing a step?

```
# podman info| grep -i runtime -A 5
  ociRuntime:
    name: crun
    package: crun-1.19.1-1.fc41.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.19.1

```

dougsland:
@slp could you please help us here? Is the patch good anything missing?

slp:
Hi @dougsland, thanks for working on this.

Even though we require the installation of `crun-krun` to install the symlink `krun -> crun`, `krun` doesn't really have an entity of its own. Calling `crun` through `krun` is just a way to let the first know it needs to use libkrun to run the container with virtualization-based isolation, but the OCI runtime is still `crun`.

On the other hand, we don't want to default to use `krun` in AutoSD. Instead, the use of `krun` should be done on selected containers, launched as quadlets, by adding `--runtime=krun` to `PodmanArgs`.

tools/qm-change-podman-runtime
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
dougsland, line 1: @slp this script is updating container config in qm.container (which we require) and also podman. Is that good solution to your eyes (only enable when running it)?

slp: I still don't think we need to change the runtime globally, but on a per-container basis.

dougsland:
@RakeshMK90 @Yarboa are working on it.

Issue #674, opened 2024-12-05
Split demos, rpms and non QM core tests to different repo


Yarboa:
QM repo contains, different use cases, which currently not related to QM core

1) Create "QM-extra" or QM-incubator, 

2) Split QM.spec file, to QM-core and QM-incubator specs

3) Move non-core issues, files, docs, tests, to "QM-incubator"

dougsland:
@Yarboa as we had many suggestions names and I created https://github.com/containers/qm-next 
The "next" refers to its role in serving as a staging area for changes that are "next in line" to be merged into the "core QM" (Example: KVM subpackage) and **likely to appear** in the next release. Also a place where it's welcome for anything related to QM, documentation, slides, demos etc. For the record, still would prefer to avoid all this changes and multiple repos, multiple CI/CD etc but not my call.

If all agree, we can start cleaning up things in the core soon. I will let this question open here.
tagging active maintainers (please add someone I missed): @Yarboa @sandrobonazzola @aesteve-rh  @kleinffm 

cc @alexlarsson

dougsland:
Hi,

Although the decision may have already been made, I just want to express my thoughts and share my prescription on this matter, and ensure they are considered:

1. **Hard-to-find code that requires long-term support:**  
   It’s challenging to identify code that will need to be maintained for the next 10 years downstream. This can be partially addressed using commands such as:
   `git log --grep="feature abc"`
   `git show -5` (the last 5 patches merged)


2. **Splitting the spec into multiple spec files:**  
This idea has full support from the QM maintainers.

3. **Reorganizing non-core content:**  
Moving documentation, tests, and presentations to separate repositories, leaving only the core components in the main repository.
*(Personally, I prefer having everything in one place—it's more convenient and feels cohesive. I'm fine with moving all demos to https://github.com/containers/demos , but in my opinion, documentation and tests should remain in the repository. That said, I fully support whatever the majority decides.)*

Let's take a broader perspective. The issue likely occurred after a change in the QM spec, which introduced three directories for subpackages, causing the autoSD installer to break. The problem was probably due to a missing '-p' flag in the 'mkdir' command within the autoSD installer.
Given this, it became challenging to 'git clone' the QM repository and identify the specific change that caused the issue, as there were numerous unrelated changes to the core QM, including subpackages, documentation, tests, presentations, and demos.

Here’s my take on this situation:  

1. The installer should be more robust and handle directory existence gracefully. It should not assume that directories will consistently either always exist or never exist. 
[...]

PR #681, opened 2024-12-30
split subpackage ros2 rolling

dougsland:


dougsland:
No answer to my point here: https://github.com/containers/qm/issues/674 so I am continuing with the split in the main repo.

PR #685, opened 2024-12-31
split subpackage tty7 into single spec and chmod +x tools

dougsland:


Yarboa (commented):
Please run this command `pre-commit run --show-diff-on-failure --color=always --all-files` please add the fixed file and commit again

I do not see here 
any additional package, created, is it ok?  
https://copr.fedorainfracloud.org/coprs/packit/containers-qm-685/packages/

dougsland:
> pre-commit run --show-diff-on-failure --color=always --all-file

yes, it's disabled by default.

dougsland:
No answer to my point here: https://github.com/containers/qm/issues/674 so I am continuing with the split in the main repo.

PR #691, opened 2025-01-01
initial code for QM python api

dougsland:


Yarboa:
@dougsland sorry, no description,  please edit PR description

Based on the new README, there is bluechictl to control containers.
Is it really needed?

Yarboa (approved):
I have no objection against this, MHO it is not needed :\

dougsland:
> I have no objection against this, MHO it is not needed :\

You are right, let's use bluechi for now, if the right use case arrives we can revive this.

PR #694, opened 2025-01-03
fix requirement of modprobe ip_tables

dougsland:
- netavark requires ip_tables, otherwise podman complains
- removed in %post isostres command (not required at all)

### DIFF
--- a/setup
+++ b/setup
@@ -151,6 +151,11 @@ install() {
         EXTRA_FLAG="--use-host-config"
     fi
 
+    # SecurityLabelNested not supported so far in CentOS Stream 9 or lower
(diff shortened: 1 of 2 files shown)


setup
```diff
@@ -151,6 +151,11 @@ install() {
         EXTRA_FLAG="--use-host-config"
     fi
 
+    # SecurityLabelNested not supported so far in CentOS Stream 9 or lower
```
Yarboa, line 154: Is it due to bootc?

dougsland: No, me and Leo found this testing CentOS @Yarboa

Issue #675, opened 2024-12-16
Add CI job to build QM minimal image using automotive-image-builder


mwperina:
To test compatibility between automotive-image-builder and QM we need to add CI to QM project, which will run automotive-image-builder to build minimal QM image using [qm.aib.yml](https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/blob/main/examples/qm.aib.yml) image with latest QM package changes

alexlarsson:
I think building https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/blob/main/examples/qm.aib.yml is better.

mkemel:
Please assign this issue to me

PR #697, opened 2025-01-05
Add test to run basic qm image build with a-i-b

mkemel:
Introducing a new sanity test, that builds a basic qm image with automotive-image-builder.

Resolves:  #675

### DIFF
--- a/tests/aib/check_qm_with_aib.sh
+++ b/tests/aib/check_qm_with_aib.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Download basic QM manifest
+curl -o qm.aib.yml "https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/raw/main/examples/qm.aib.yml?ref_type=heads"
+
+USE_QM_COPR="${PACKIT_COPR_PROJECT:-rhcontainerbot/qm}"
+COPR_URL="https://download.copr.fedorainfracloud.org/results/${USE_QM_COPR}/epel-9-$(uname -m)/"
+#shellcheck disable=SC2089
+EXTRA_REPOS='extra_repos=[{"id":"qm_build","baseurl":"'"$COPR_URL"'"}]'
+
+# Run AIB in container
+curl -o auto-image-builder.sh "https://gitlab.com/CentOS/automotive/sample-images/-/raw/main/auto-image-builder.sh?ref_type=heads"
+#shellcheck disable=SC2027,SC2090,SC2086
+/bin/bash auto-image-builder.sh build --export qcow2 --define "'"$EXTRA_REPOS"'" qm.aib.yml qm.qcow2
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -113,6 +113,18 @@ jobs:
             virtualization:
              is-supported: true
 
+  - job: tests
+    trigger: pull_request
+    identifier: automotive-image-builder
+    tmt_plan: /plans/e2e/aib
+    targets:
+      - epel-9-x86_64
+    tf_extra_params:
+      environments:
+          hardware:
+            disk:
+             - size: ">= 20 GB"
+
 
   - job: propose_downstream
     trigger: release
--- a/plans/e2e/aib.fmf
+++ b/plans/e2e/aib.fmf
@@ -1,0 +1,11 @@
+summary: automotive-image-builder QM build test
+
+discover:
+    how: fmf
+    filter: 'tag:aib'
+
+execute:
+    how: tmt
+
+report:
+    how: junit
--- a/tests/aib/check_qm_with_aib.fmf
+++ b/tests/aib/check_qm_with_aib.fmf
@@ -1,0 +1,6 @@
+summary: Test basic QM image build with automotive-image-builder
+test: /bin/bash ./check_qm_with_aib.sh
+duration: 20m
+tag: aib
+framework: shell
+id: 146b987f-698f-4f84-8029-eaa253d049ce


dougsland:
looks sane to me @Yarboa PTAL

Yarboa:
@mkemel 
AFAICT, I see that no gate run this test.
Few questions here:
- In general the test itself just build qm qcow image. is it?
- Does it check that vm also running? 
- Do you have dependency in /dev/kvm? 
- Dont you think it should be part of automotive-sig? 

One more thing here, What the tests should really tests?
Is it a new QM image in autosd image with latest qm image?

Yarboa (commented):
please refer this
https://github.com/containers/qm/pull/697#issuecomment-2572597949

mkemel:
@alexlarsson care to take a look?

Yarboa:
> Hi @Yarboa
> 
> 1. Yes
> 2. No
> 3. Not that I'm aware of, don't think that osbuild has such a dependency.
As per this https://github.com/osbuild/osbuild?tab=readme-ov-file#requirements
only qemu-imag, ack

> 4. The test should verify that a PR does not break QM compatibility with a-i-b and the build runs successfully. So it should be in QM.
So i thing there is an issue here, I will review PR again

> 
> One question from my side is - the requirement is for CI only. So maybe it would be better to create a separate plan for that?

The issue here is not the plan it is packit,
I would creatye tests with special hardware in pacvklit.yaml

mkemel:
> The issue here is not the plan it is packit,
I would creatye tests with special hardware in pacvklit.yaml

Currently the test is run by packit as part of this job:
```
  - job: tests
    trigger: pull_request
    identifier: qm-tier-0
    tmt_plan: /plans/e2e/tier-0
    targets:
      - epel-9-x86_64
    tf_extra_params:
      environments:
        - artifacts:
          - *bluechi_copr_repo
          hardware:
            disk:
             - size: ">= 20 GB"
```
And the only hardware requirement is `>= 20GB`, and the test does not require that. But do we want this test have its own job?

tests/qm-sanity-test/check_qm_with_aib.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Create AIB manifest with the correct QM copr
+USE_QM_COPR="${PACKIT_COPR_PROJECT:-rhcontainerbot/qm}"
+sed -e "s,\${QM_COPR},${USE_QM_COPR},g" check_qm_with_aib.aib.yml > qm.aib.yml
+
+AIB="automotive-image-builder build --export qcow2 qm.aib.yml qm.qcow2"
+BUILDDIR=_build
```
michalskrivanek, line 8: from here on, it would be best to use the auto-image-builder.sh instead of copy&paste

michalskrivanek (changes requested):

tests/qm-sanity-test/check_qm_with_aib.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Create AIB manifest with the correct QM copr
+USE_QM_COPR="${PACKIT_COPR_PROJECT:-rhcontainerbot/qm}"
+sed -e "s,\${QM_COPR},${USE_QM_COPR},g" check_qm_with_aib.aib.yml > qm.aib.yml
+
+AIB="automotive-image-builder build --export qcow2 qm.aib.yml qm.qcow2"
```
Yarboa, line 7: Can you explain how this file only helps to understand which mpp to use? 
I do not see use_qm true or something which this repo installed should be read?

tests/qm-sanity-test/check_qm_with_aib.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Create AIB manifest with the correct QM copr
+USE_QM_COPR="${PACKIT_COPR_PROJECT:-rhcontainerbot/qm}"
```
Yarboa, line 4: OK that looks good
https://download.copr.fedorainfracloud.org/results/packit/containers-qm-685/epel-9-$basearch/

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: This is the new simplified manifest format
https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/merge_requests/130

As for the repo, it is specified in `qm.aib.yml` which is produced from `check_qm_with_aib.aib.yml` intorduced here above.

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: Exactly

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: But then I would need to download it from gitlab? We don't have an EPEL package for `sample-images`, do we?

Yarboa:
> tmt_plan: /plans/e2e/tier-0

I do not think it is related, 
All tests running are running on QM environment, and its prepare step, set QM env
Current test, have no relation with qm deployments, just verify qm image is build  success with qm PR generated.

Isnt it?

tests/qm-sanity-test/check_qm_with_aib.sh
michalskrivanek: correct. but it's still better to use that instead of this complex cmdline.
but yeah, we though about moving it into a-i-b repo instead...

mkemel:
> I do not think it is related,
All tests running are running on QM environment, and its prepare step, set QM env
Current test, have no relation with qm deployments, just verify qm image is build success with qm rpm PR generated.

Yes. I wasn't sure if I should just add this one to sanity tests (this is a sanity test after all), and run it in an environment with QM (which does not affect this test), or create a separate plan and run it as a separate job. I chose the former in avoid creating separate plan and job for one test. But if you think that it would be better to have a separate job for this one - I can add one, it's not a problem.

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: So I considered cloning the `sample-images` repo and run `auto-image-builder.sh` from there, but I thought that using the script from upstream/main might be unstable, and would add git installtion dependency. On the other hand, the beauty of having a container image with everything needed inside is that you don't really need to set anything up, and it runs with one command in (almost) any environment with podman. So actually I like one complex cmdline more than cloning a whole repo just to get the arguments straight. WDYT?

mkemel: Maybe putting `auto-image-builder.sh` in the container image would make sense? This would somewhat simplify this one-liner

Yarboa:
> > I do not think it is related,
> > All tests running are running on QM environment, and its prepare step, set QM env
> > Current test, have no relation with qm deployments, just verify qm image is build success with qm rpm PR generated.
> 
> Yes. I wasn't sure if I should just add this one to sanity tests (this is a sanity test after all), and run it in an environment with QM (which does not affect this test), or create a separate plan and run it as a separate job. I chose the former in avoid creating separate plan and job for one test. But if you think that it would be better to have a separate job for this one - I can add one, it's not a problem.

Ack, ask for different machine with specific plan+test running, there is no prepare step here,
it needs only c9s machine with minimum of memory and storage, in case of no nested kvm, no need to ask it from testing farm

tests/qm-sanity-test/check_qm_with_aib.sh
Yarboa: Please refer this issue
#675 now, it is more clear, so there is sample manifest and additional support in code https://gitlab.com/CentOS/automotive/src/automotive-image-builder/-/commit/f28b7f240a9a33f2273c2d8a578b3ebd4aad6623

simplifeid manifest format

tests/qm-sanity-test/check_qm_with_aib.sh
Yarboa: Ok the resolve added in description, info clear the picture
I think a readme section should be added here, 

https://github.com/containers/qm/tree/main/docs/devel#building-centos-autosd-and-qm-manually

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: > which container do you want to include it in? qm?
No, in `quay.io/centos-sig-automotive/automotive-osbuild`

Yarboa:
> > The issue here is not the plan it is packit,
> > I would creatye tests with special hardware in pacvklit.yaml
> 
> Currently the test is run by packit as part of this job:
> 
> ```
>   - job: tests
>     trigger: pull_request
>     identifier: qm-tier-0
>     tmt_plan: /plans/e2e/tier-0
>     targets:
>       - epel-9-x86_64
>     tf_extra_params:
>       environments:
>         - artifacts:
>           - *bluechi_copr_repo
>           hardware:
>             disk:
>              - size: ">= 20 GB"
> ```
> 
> And the only hardware requirement is `>= 20GB`, and the test does not require more than that. But do we want this test have its own job?

better to maintain it separately in case it is broken, gate run also on bare c9s/fedora, it should not block merge of non related  PRs

tests/qm-sanity-test/check_qm_with_aib.sh
michalskrivanek: a bad merge to main is exactly what you want to detect anyway, isn't it?
curl -O https://gitlab.com/CentOS/automotive/sample-images/-/raw/main/auto-image-builder.sh?ref_type=heads would work...

dougsland:
Worth add for reference this comment and discussion: https://github.com/containers/qm/issues/674#issuecomment-2532089601

dougsland:
If @michalskrivanek @Yarboa are satisfied we can proceed. I trust in their review.

Yarboa:
> If @michalskrivanek @Yarboa are satisfied we can proceed. I trust in their review.

I thougt there is agreement on that 
https://github.com/containers/qm/pull/697#issuecomment-2574931376, waiting for another update

BTW, it could be moved to qm-next also

michalskrivanek:
well, I'm NOT happy with copying the logic from auto-image-builder.sh. It even changed in the meantime...

tests/qm-sanity-test/check_qm_with_aib.sh
mkemel: Done.
Took some time as I was in some massive virus knockout

mkemel:
1. Created separate plan and Packit job to run it
2. Now use auto-image-builder.sh

Yarboa (approved):
LGTM

tests/aib/check_qm_with_aib.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Create AIB manifest with the correct QM copr
+USE_QM_COPR="${PACKIT_COPR_PROJECT:-rhcontainerbot/qm}"
+sed -e "s,\${QM_COPR},${USE_QM_COPR},g" check_qm_with_aib.aib.yml > qm.aib.yml
```
michalskrivanek, line 5: maybe one more thing...why doing it in such an ugly way? you can pass a variable by --define, or add extra_repo...see the example in https://sigs.centos.org/automotive/building/building_an_os_image_that%20uses_a_custom_kernel/#build-an-autosd-image-with-a-custom-kernel-and-run-it-on-physical-hardware

mkemel: Good point. Done

michalskrivanek (approved):

Yarboa:
@mkemel  please check precommit errors
https://github.com/containers/qm/actions/runs/12790524518/job/35695989011?pr=697

mkemel:
@Yarboa It took me some time to make this work, the whole quotes thing in this script is quite complicated since we want to pass a mix of single and double quotes to a shell script which passes it to podman which passes it to container. Suppressed the warnings, shellcheck does not really recognize what I'm doing there.

Yarboa:
> @Yarboa It took me some time to make this work, the whole quotes thing in this script is quite complicated since we want to pass a mix of single and double quotes to a shell script which passes it to podman which passes it to container. Suppressed the warnings, shellcheck does not really recognize what I'm doing there.

Did you try this one? 
`EXTRA_REPOS="extra_repos=[{\"id\":\"qm_build\",\"baseurl\":\"$COPR_URL\"}]"`

And that one?
`/bin/bash auto-image-builder.sh build --export qcow2 --define "$EXTRA_REPOS" qm.aib.yml qm.qcow2`

mkemel:
> Did you try this one?
EXTRA_REPOS="extra_repos=[{\"id\":\"qm_build\",\"baseurl\":\"$COPR_URL\"}]"

This would work, but I like it less tbh, I think using single quote is much better than escaping every double quote, it makes reading much more difficult to my taste

>
And that one?
/bin/bash auto-image-builder.sh build --export qcow2 --define "$EXTRA_REPOS" qm.aib.yml qm.qcow2

This wouldn't work because we need single quotes there

PR #707, opened 2025-01-20
Add jq package installation to a-i-b test

mkemel:
auto-image-builder.sh now uses jq. Adding jq to the prepare step for automotive-image-builder test plan


mkemel:
@dougsland @Yarboa PTAL

PR #709, opened 2025-01-21
Makefile: make rpm should only generate qm

dougsland:


dougsland:
@Yarboa going to merge this one but I know you are working in a general improvement. This one is just to unblock build failures when using Containerfile git clone the repo.

Issue #384, opened 2024-04-24
docs: qm.container the meaning of this file


dougsland:
qm.container contain multiple definitions that explain what QM is behind the scenes. Explore this file a bit in the README.md.

dougsland:
Here an example how we documented options in the past: https://github.com/containers/qm/blob/main/qm.container#L12-L16

dougsland:
@pengshanyu should we close this one?

PR #710, opened 2025-01-27
add comments to qm.container service

pengshanyu:
resolve part of: https://github.com/containers/qm/issues/384
Update [Service] section in qm.container to describe in short comments what are the options.


pengshanyu:
I am not sure about the parameter values for qm scenario, I did some searching and refer to "systemd.resource-control - Resource control unit settings".
Hi @dougsland , @Yarboa , could you please give me some review and reference? Thanks.

Yarboa:
There is also this PR #678
This is quadlet file, it is all written in man quadlet

related to man please search for /\[Service\]
It refers you to that one 

```
[Service] table and [Install] tables pass directly to systemd and are handled by  it.
       See systemd.unit(5) man page for more information.
```

So I would suggest to show where to fine the relevant directive for the service
man systemd.directives
You could find each directive, to refer 
```
man systemd.directives | grep -A1 ".* AllowedCPUs=.*"
       AllowedCPUs=
           systemd.resource-control(5)

```
then man systemd.resource-control explains the relevant information

Please share your thoughts here

pengshanyu:
Hi @Yarboa , thank you so much for your suggestion. 
You are right. Refer to the relevant man section people can get more accurate information.
You mean that in the comment, we just need to write the man command for example `man systemd.resource-control`, right? So that they can directly find the relevant parameters.

qm.container
```diff
@@ -4,6 +4,24 @@ WantedBy=default.target
 [Service]
 # It's recommended to use systemd drop-in to override the
 # systemd settings. See QM manpage for an example.
+# 1) CPUWeight: This setting controls the CPU controller for qm.
```
Yarboa, line 7: In reply to your question
I would consider this:
QM cgroup, pass directly to systemd and handled by it, refer
man systemd.resource-control for detailed resource-control

Yarboa (commented):
please see my comment above

qm.container
dougsland: +1

pengshanyu: Just came back.
I improved it according to the comment, thank you all for the details.

Yarboa (approved):
LGTM

PR #711, opened 2025-01-30
rename /usr/share/qm/seccomp.json

dougsland:
To be explicit the current seccomp.json
DO NOT support realtime (rt) sched we will
add it into the name of the file.


dougsland:
@alexlarsson as we talked.

PR #712, opened 2025-01-30
README: add a note about how to enable realtime

dougsland:


dougsland:
cc @alexlarsson

README.md
```diff
@@ -587,6 +588,14 @@ Looking for quadlet examples files? See our [docs dir](docs/quadlet-examples/).
 
 If your looking for contribute to the project use our [development README guide](docs/devel/README.md) as start point.
 
+## Realtime
+
+To enable real-time removal of sched_* blockage via seccomp, comment out the SeccompProfile.
+
+```bash
```
Yarboa, line 595: IT should be qm-container,
And it should be in quadlet dropin with ovveride 
this one stays as it is but the dropin should have the following 

[Container]
SeccompProfile=""

maybe something like that

```
cat << EOF >> /etc/containers/systemd/qm.container.d/rt.conf
> [Container]
SeccompProfile=""
> EOF
```

Yarboa, line 595: IT should be qm-container,
And it should be in quadlet dropin with ovveride 
this one stays as it is but the dropin should have the following 

[Container]
SeccompProfile=""

maybe something like that

```
cat << EOF >> /etc/containers/systemd/qm.container.d/rt.conf
> [Container]
SeccompProfile=""
> EOF
```

dougsland:
bandit-check failing not related to the change.

Yarboa:
> bandit-check failing not related to the change.

Pleasee this in bandit check
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

PR #718, opened 2025-02-10
qm.if: allow qm process to audit and ptrace

dougsland:
## Summary by Sourcery

Chores:
- Update qm.if.

### DIFF
--- a/qm.if
+++ b/qm.if
@@ -580,6 +580,24 @@ template(`qm_domain_template',`
 	')
 ')
 
+# Allow qm_t domain to read auditd.service
+interface(`qm_allow_auditd_read',`
+    gen_require(`
+        type qm_t, auditd_unit_file_t;
+    ')
+
+    allow qm_t auditd_unit_file_t:file read;
+')
+
+# Allow qm_t domain to ptrace other qm_t processes
+interface(`qm_allow_ptrace_qm',`
+    gen_require(`
+        type qm_t;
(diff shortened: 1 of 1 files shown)


qm.if
```diff
@@ -580,6 +580,24 @@ template(`qm_domain_template',`
 	')
 ')
 
+# Allow qm_t domain to read auditd.service
+interface(`qm_allow_auditd_read',`
```
rhatdan, line 584: NO idea where this one came from?

qm.if
```diff
@@ -580,6 +580,24 @@ template(`qm_domain_template',`
...
+    allow qm_t auditd_unit_file_t:file read;
+')
+
+# Allow qm_t domain to ptrace other qm_t processes
+interface(`qm_allow_ptrace_qm',`
+    gen_require(`
+        type qm_t;
+    ')
+
+    allow qm_t qm_t:process ptrace;
```
rhatdan, line 598: Do you also want to allow qm_t to ptrace container processes within the QM?

If you really need this, then it should be disabled by default and then wrapped in a boolean to allow it only for short periods of time. I would much prefer to run debuggers and ptrace type commands in the ASIL (unconfined_t) domain then inside of the qm.  An ASIL process can ptrace a qm process.

dougsland: @rhatdan perfect, closing.

Issue #723, opened 2025-02-13
qm.container mounts /var/tmp as tmpfs


alexlarsson:
qm.container has `ReadOnly=true`, which maps to `podman run --read-only`.
This tells podman to mount the container image read-only, and then mount a tmpfs on top of /tmp, /run and /var/tmp.

However, in QM, /var/tmp is generally on top of the /var volume (which comes from /var/qm on the host), and we *don't* want this part to be a tmpfs, because that limits the size of /var/tmp files in the qm partition, and this is not expected (e.g. it is different than on what normally happens in rhel, where /tmp is ram and /var/tmp is disk based).

I think we need to pass some extra argument to manually configure /tmp and /run as tmpfs, with not special options for /var/tmp (so its just picked up from the volume).

I think what we want is
ReadOnly=true
ReadOnlyTmpfs=false
Mount=type=tmpfs,tmpfs-size=512M,destination=/tmp
Mount=type=tmpfs,tmpfs-size=512M,destination=/run

alexlarsson:
@rhatdan does this make sense to you?

dougsland:
@rhatdan @alexlarsson  I asked QE to test this.

dougsland:
See-Also: https://github.com/containers/qm/issues/722

pengshanyu:
I verified that the following configuration works:
```
ReadOnly=true
ReadOnlyTmpfs=false
Mount=type=tmpfs,tmpfs-size=512M,destination=/tmp
Mount=type=tmpfs,tmpfs-size=512M,destination=/run
Mount=type=tmpfs,destination=/dev/shm
```

dougsland:
> Mount=type=tmpfs,destination=/dev/shm

thanks a lot.

PR #727, opened 2025-02-21
qm.container: add tmpfs additional flags

dougsland:
Resolves: https://github.com/containers/qm/issues/723

## Summary by Sourcery

New Features:
- Adds support for specifying additional flags when mounting a tmpfs volume in a container.

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -80,11 +80,18 @@ Network=private
 # Makes the container's filesystem read-only, enhancing security by preventing modifications.
 ReadOnly=true
 
+# TmpFS flags
+ReadOnlyTmpfs=false
+Mount=type=tmpfs,tmpfs-size=512M,destination=/tmp
+Mount=type=tmpfs,tmpfs-size=512M,destination=/run
+Mount=type=tmpfs,destination=/dev/shm
+
 # Rootfs
 # ------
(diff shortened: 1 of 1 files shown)


qm.container
```diff
@@ -80,11 +80,19 @@ Network=private
 # Makes the container's filesystem read-only, enhancing security by preventing modifications.
 ReadOnly=true
 
+# TmpFS flags
+ReadOnlyTmpfs=false
+Mount=type=tmpfs,tmpfs-size=512M,destination=/tmp
+Mount=type=tmpfs,tmpfs-size=512M,destination=/run
+
```
pengshanyu, line 87: I have tested it, we also need to add the following code:
`Mount=type=tmpfs,destination=/dev/shm`
otherwise the following issue will occur:
`Error: failed to get new shm lock manager: failed to create 2048 locks in /libpod_lock: read-only file system`

Yarboa (commented):
Please remove that lines from ostree testing 
QM_TMP_DIR="/var/tmp.dir"
We will have a clue that this fix is working, 
https://github.com/containers/qm/blob/main/tests/ffi/common/prepare.sh#L75-L79

BTW all that make 
subsystems/img_tempdir could be deleted.
and alsodeleted from READMEs

Issue #722, opened 2025-02-13
improve /ffi/disk test


pengshanyu:
The code from [L8](https://github.com/containers/qm/blob/main/tests/ffi/disk/test.sh#L8) to [L12](https://github.com/containers/qm/blob/main/tests/ffi/disk/test.sh#L12) should be improved. During qm setup, when qm create additional partition, it create "part /var"

dougsland:
@pengshanyu could you please verify: https://github.com/containers/qm/pull/727 ?

pengshanyu:
Sure.

dougsland:
@pengshanyu can we close this one with recent qm-0.7.2 release?

pengshanyu:
Actually https://github.com/containers/qm/pull/740 resolve this issue.

Yarboa:
IT seems that timeouts happens in that tests
740 is failing

dougsland:
> Actually [#740](https://github.com/containers/qm/pull/740) resolve this issue.

Okay. Closing to avoid dup.

Issue #721, opened 2025-02-11
CI/CD checks fails on agent-flood.


nsednev:
During basic CI/CD checks I see that Packit-as-a-Service / testing-farm:centos-stream-9-x86_64:e2e-ffi  fails on /tests/ffi/agent-flood.
Here goes some Testing Farm [artifacts](https://artifacts.dev.testing-farm.io/4b6ca6d7-a95c-440d-8ae9-7f82693ca91c/) and 
[Dashboard](https://artifacts.dev.testing-farm.io/4b6ca6d7-a95c-440d-8ae9-7f82693ca91c/).

WARN[0010] StopSignal SIGTERM failed to stop container ffi-qm in 10 seconds, resorting to SIGKILL 
Deleted: c2febf90cc5baa7cfbdcecc1e485300cd110c0c0bb9b39fd3f39564d8916e330
[ INFO  ] PASS: Command podman exec -it qm /bin/bash -c "podman  rmi -f --all" successful
[ INFO  ] PASS: Command rm -rf /var/qm/lib/containers/registry/tools-ffi:latest/b8e9f060dd20af7fdd2e577300d219b18aefc753538f18b603885a6d0945cb29 successful
[ INFO  ] PASS: Command rm -rf /etc/containers/systemd/qm.container.d/ successful
[ INFO  ] PASS: Command systemctl daemon-reload successful
[ INFO  ] PASS: Command systemctl restart qm successful
[ INFO  ] PASS: Command rm -f  successful
[ INFO  ] PASS: Command mkdir -p /etc/containers/systemd/qm.container.d/ successful
[ INFO  ] PASS: Command systemctl daemon-reload successful
[ INFO  ] PASS: Command systemctl restart qm successful
The system is not running under OSTree.
Trying to pull quay.io/centos-sig-automotive/ffi-tools:latest...
Getting image source signatures
Copying blob sha256:972d4b520f080b2d880670a56c847c9e42ad7a01207b43066f965c69a04a6d91
Copying blob sha256:376ee6a8ddb9979b92f63edbb9a11f5f26ef64580d37a54cd9edb27dc5842f98
Copying config sha256:c2febf90cc5baa7cfbdcecc1e485300cd110c0c0bb9b39fd3f39564d8916e330
Writing manifest to image destination
c2febf90cc5baa7cfbdcecc1e485300cd110c0c0bb9b39fd3f39564d8916e330
[...]

PR #740, opened 2025-03-03
improve the name of var partition

pengshanyu:
Resolve: https://github.com/containers/qm/issues/722

## Summary by Sourcery

Tests:
- Improve the logic to detect the /var partition in the test script.


tests/ffi/disk/test.sh
```diff
@@ -5,17 +5,25 @@
...
 check_var_partition(){
+   var_partition_name="part /var/qm"
+
    if stat /run/ostree-booted > /dev/null 2>&1; then
-      qm_var_partition="part /var"
+      var_partition_name="part /var"
    else
-      qm_var_partition="part /usr/lib/qm/rootfs/var"
+      local release_id
+      release_id=$(grep -oP '(?<=^ID=)\w+' <<< "$(tr -d '"' < /etc/os-release)")
```
Yarboa, line 14: consider using this instead 
podman exec qm findmnt /var

BYW, which distro cause that test to be unstable?

pengshanyu: It is not unstable, we have a separate `part /var/qm` partition in the image.

pengshanyu: I tried with `podman exec qm findmnt /var`, it seems that it cannot be used as a criterion to determine whether a separate /var partition exists, because in the image without a separate /var partition, this command still succeeds.

pengshanyu:
I'm not sure why there are failures of `SRPM build failed`/`copying system image from manifest list: parsing image configuration: fetching blob: received unexpected HTTP status: 502 Bad Gateway   INFO: Pulling image: registry.fedoraproject.org/fedora:41`
Hi @Yarboa @dougsland , could you help to take a look, thanks.

tests/ffi/disk/test.sh
```diff
@@ -5,17 +5,29 @@
 . ../common/prepare.sh
 
 check_var_partition(){
+   name_of_qm_var_partition="part /var/qm"
+   seccomp_file_name="/usr/share/qm/seccomp.json"
```
Yarboa, line 9: Consider querying it only if it exist in PodmanArgs.
In case no PodmanArgs, at all, I assume you should use container Memory instead

tests/ffi/disk/test.sh
```diff
@@ -33,18 +45,14 @@ OOMScoreAdjust=1000
 
 [Container]
 PodmanArgs=
-PodmanArgs=--pids-limit=-1 --security-opt seccomp=/usr/share/qm/seccomp-no-rt.json --security-opt label=nested --security-opt unmask=all --memory 5G
+PodmanArgs=--pids-limit=-1 --security-opt seccomp=${seccomp_file_name} --security-opt label=nested --security-opt unmask=all --memory 5G
```
Yarboa, line 48: In case there is not PodmanArgs, see above
You should only use Memory

tests/ffi/disk/test.sh
pengshanyu: ack

tests/ffi/disk/test.sh
pengshanyu: done

Yarboa:
LGTM, i see this one #721 becomes urgent

pbrilla-rh:
@Yarboa @dougsland can this test rework be merged?

Yarboa:
> @Yarboa @dougsland can this test rework be merged?

@pbrilla-rh thereis an open question, here
https://github.com/containers/qm/pull/740#issuecomment-2708718051

pengshanyu:
> > @Yarboa @dougsland can this test rework be merged?
> 
> @pbrilla-rh thereis an open question, here [#740 (comment)](https://github.com/containers/qm/pull/740#issuecomment-2708718051)

removed [L75-79](https://github.com/containers/qm/blob/main/tests/ffi/common/prepare.sh#L75-L79)

PR #746, opened 2025-03-05
Added yaml prefix subpackages build

Yarboa:
_Fix: GitHub Actions & QM Subpackage Build Process_

Another issue raised, #755 755

This PR fixes a long-standing GitHub Actions issue and refactors the QM subpackage build process for better maintainability.

Changes:
- Fix GitHub Actions, which was broken since day one.
- Remove img_temp subpackage (#747) & tmp_image (Handled in #727).
- Refactor build process:
- Use individual spec files for subpackages.
- Add a dedicated make target for building subpackages.
- Update documentation.
- This simplifies the build process and improves maintainability.


rhatdan:
LGTM
@dougsland @alexlarsson PTAL

rpm/kvm/qm-kvm.spec
```diff
@@ -4,25 +4,15 @@
 %global rootfs_qm %{_prefix}/lib/qm/rootfs/
 
 Name: qm-kvm
-Version: 0
+Version: 0.7.3
```
dougsland, line 7: lets keep 0 for automation

rpm/text2speech/text2speech.spec
```diff
@@ -4,26 +4,17 @@
 %global rootfs_qm %{_prefix}/lib/qm/rootfs/
 
 Name: qm-text2speech
-Version: 0
+Version: 0.7.3
```
dougsland, line 7: same here?

Yarboa: Ack

rpm/kvm/qm-kvm.spec
Yarboa: Ack.

.github/workflows/check-subpackages.yml
```diff
@@ -6,12 +6,16 @@ on:
 jobs:
   build-subpackages:
     runs-on: ubuntu-latest
```
nsednev, line 8: Why runs on ubuntu and not on fedora?

Yarboa: since it is githubactions, which run only on ubuntu IIRC
this is why we use container in it 
see this:
    container:
      image: fedora:latest  # Use Fedora as the container image

rpm/kvm/qm-kvm.spec
```diff
@@ -4,25 +4,15 @@
...
 BuildArch: noarch
 
-# Build requirements
-BuildRequires: make
-BuildRequires: podman
-BuildRequires: selinux-policy
-BuildRequires: selinux-policy-devel
-
-# Runtime dependencies
 Requires: qm = %{version}-%{release}
```
Yarboa, line 15: @dougsland this is tricky
@raballew tried to install it on AutoSD and had dependency issues, I do not think it is a must, 
Can we dismiss this dependency?

contyk: As we've discovered, this dependency is too strict to be practical.  `%{release}` includes `%{dist}` defined by the system, which might differ from `%{dist}` defined by the buildsystem used to build qm.  Furthermore, this potentially prevents qm from being updated as it would result in broken dependencies.

Relax this to `Requires: qm >= %{version}` to avoid these... challenges.

Yarboa: Ack

dougsland: >= works?

dougsland: or simply remove for now to unlock @raballew

Yarboa: just checking

PR #315, opened 2024-01-18
Gate run requires epel targets

Yarboa:
With introduction of vsomeip-selinux
as build dependency c9s target build is failing

resolve #299

PR #748, opened 2025-03-14
Fix suport for CPUWeight

alexlarsson:
In the current setup, qm.service is part of the QM.slice. This means that the CPUWeight of qm.service is only applied against the other groups of QM.slice, i.e. none. This means CPUWeight does nothing.

Here change the parent of qm.service to be -.service (the root cgroup), which means that the weight of it will be compared to the other toplevel slices, which normally is user.slice and system.slice.

So, with the current default CPUWeight in qm.container of 50, and the default weight of 100, this means that qm.service now will get half the cpu of the processes in system.slice, which is the intent.

I tried this by running "openssl speed --seconds 99999" both inside qm and outside it one a one-cpu VM, and before both processes always got 50% cpu, but now they are split up according to CPUWeight in the qm.container.

## Summary by Sourcery

Bug Fixes:
- CPUWeight of qm.service was not being applied correctly because it was only being compared against other groups in QM.slice.

### DIFF
--- a/demos/demo
+++ b/demos/demo
@@ -105,20 +105,15 @@ status() {
 }
 
 cpuweight() {
-    exec_color "sudo systemctl set-property --runtime QM.slice CPUWeight=50"
     exec_color "sudo systemctl set-property --runtime qm.service CPUWeight=50"
 
-    echo "Value stored in QM.slice/cpu.weight:"
-    sudo cat /sys/fs/cgroup/QM.slice/cpu.weight
+    echo "Value stored in qm.service/cpu.weight:"
+    sudo cat /sys/fs/cgroup/qm.service/cpu.weight
 
-    echo "Value stored in QM.slice/qm.service/cpu.weight:"
-    sudo cat /sys/fs/cgroup/QM.slice/qm.service/cpu.weight
-    exec_color "sudo systemctl set-property --runtime QM.slice CPUWeight=10"
+    exec_color "sudo systemctl set-property --runtime qm.service CPUWeight=10"
 
-    echo "Value stored in QM.slice/cpu.weight:"
-    sudo cat /sys/fs/cgroup/QM.slice/cpu.weight
-    echo "Value stored in QM.slice/qm.service/cpu.weight:"
-    sudo cat /sys/fs/cgroup/QM.slice/qm.service/cpu.weight
+    echo "Value stored in qm.service/cpu.weight:"
+    sudo cat /sys/fs/cgroup/qm.service/cpu.weight
     echo -e "\n\n[Press enter to continue]"
     read -r
 }
--- a/docs/devel/README.md
+++ b/docs/devel/README.md
@@ -156,7 +156,7 @@ ago
       Tasks: 7 (limit: 7772)
      Memory: 82.1M (swap max: 0B)
         CPU: 945ms
-     CGroup: /QM.slice/qm.service
+     CGroup: /qm.service
              ├─libpod-payload-a83253ae278d7394cb38e975535590d71de90a41157b547040
 4abd6311fd8cca
              │ ├─init.scope
--- a/qm.8.md
+++ b/qm.8.md
@@ -40,7 +40,7 @@ systemctl status qm.service
       Tasks: 11 (limit: 76801)
      Memory: 275.1M (swap max: 0B)
         CPU: 4.527s
-     CGroup: /QM.slice/qm.service
+     CGroup: /qm.service
              ├─libpod-payload-00de006493bc970788d6c830beb494a58a9a2847a5eda200812d3a8b4e214814
              │ ├─init.scope
              │ │ └─993676 /sbin/init
@@ -53,11 +53,13 @@ systemctl status qm.service
 ...
 ```
 
-## CGROUPS QM.slice
+## CGroups and container configuration
 
-Notice that the QM environment is running systemd and other services within the
-QM.Slice. This slice can be used to modify the cgroups controls of all of the
-processes within the QM environment.
+The options in the qm.container file overridden by using drop-in files, in the
+directories `/etc/containers/systemd/qm.container.d` or`
+`/usr/lib/containers/systemd/qm.container.d. This allows overriding for example
+CGroup options like Service.CPUWeight, or podman options like Container.Volume.
+Such options will affect all the processes running in the qm container.
 
 ## Install Additional packages in QM
 
--- a/qm.container
+++ b/qm.container
@@ -34,7 +34,10 @@ MemorySwapMax=0
 # Containers within the qm contain default set OOMScoreAdj to 750
 OOMScoreAdjust=500
 Restart=always
-Slice=QM.slice
+# qm.service is a toplevel cgroup, so CPUWeight is relative to all other cgroups in that
+# parent (such as user.slice and system.slice), otherwise the CPUWeight of qm.service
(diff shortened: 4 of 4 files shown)


engelmi (approved):
LGTM

rhatdan:
@pypingou @jtligon  this might be the source of the problem the customer was reporting.

rhatdan:
LGTM

jtligon:
It might be, we'll take a look. tyvm sir!

dustinblack:
I'm not convinced it is the best approach to move the `qm.service` outside of the `QM.slice` hierarchy. Better, I think, would be to apply the constraints we are interested in directly to the slice. I would keep the existing `Slice=QM.slice` here and add the resource controls we want to a new `/etc/systemd/system.control/QM.slice.d/50-resource_controls.conf` file.

Yarboa:
> /etc/systemd/system.control/QM.slice.d/50-resource_controls.conf

@alexlarsson 
There is some logic here, isnt it?  

create /etc/systemd/system/QM.slice:
[Slice]
CPUWeight=50

And the whole service will run inside this QM.slice
As per `man systemd.slice` 
```
A unit configuration file whose name ends in ".slice" encodes information about a slice unit. A slice
       unit is a concept for hierarchically managing resources of a group of processes. 
```

It also mentions that 
```
See systemd.unit(5) for the common options of all unit configuration files
```

So i assume we should check configuration as following, got the relevant configuration, 
```
[root@localhost ~]# systemctl show QM.slice | grep CPUWe
CPUWeight=[not set]
StartupCPUWeight=[not set]
```

Is it correct? @dustinblack

alexlarsson:
> I'm not convinced it is the best approach to move the `qm.service` outside of the `QM.slice` hierarchy. Better, I think, would be to apply the constraints we are interested in directly to the slice. I would keep the existing `Slice=QM.slice` here and add the resource controls we want to a new `/etc/systemd/system.control/QM.slice.d/50-resource_controls.conf` file.

I disagree, the QM.slice cgroup is completely useless (it has only ever one child) and was added with no actual reason. Making configuration of qm more complicated by arbitrary splitting in what files various different options are specified makes it harder to use and document and has zero advantages.

If anyone can give any reason at all for the existance of QM.slice, please do so. It was added by dan and me with no real thought behind it in the early qm work.

Yarboa:
> I disagree, the QM.slice cgroup is completely useless (it has only ever one child) and was added with no actual reason. Making configuration of qm more complicated by arbitrary splitting in what files various different options are specified makes it harder to use and document and has zero advantages.
> 
> If anyone can give any reason at all for the existance of QM.slice, please do so. It was added by dan and me with no real thought behind it in the early qm work.

Sure, it is good to have docs, 
https://github.com/containers/qm/blob/main/qm.8.md#cgroups-qmslice
As mentioned it is good for controlling subslices

And i found in other place 

```
By creating a slice, the ASIL environment can easily modify the cgroups of all processes running within the QM container environment.
```

so, If IIUC correctly, you are recommending maintaining QM.slice
But recommend to move running QM service under .slice instead of QM.slice

And the cgroups will  look like that 
Control group /:
-.slice
├─qm.service … (#84482)

Instead of 
Control group /:
-.slice
├─QM.slice (#315)
│ └─qm.service … (#86141)

Correct?

dustinblack:
I don't have a strong technical argument for it, but it still feels to me like the QM.slice is the right way to go. I find it semantically strange to have a service directly under the -.slice since all other services I believe belong to either the system.slice or the user slice. We also can't be sure of future architectural needs of the QM, and it seems like starting with the standard of a slice that may be subdivided as needed is more prudent.

We should also account for the fact that some not insignificant amount of existing documentation and user communication includes the QM.slice, so if we go forward with this change we need acknowledge that it is an architectural deviation and we must have the due diligence to communicate that and ensure docs and diagrams are updated.

alexlarsson:
All the current options what we change, including he current value of CPUWeight (which doesn't work) are in the qm.container file, and any changes to this file will affect *all* qm processes (at least if we drop qm.slice). The later means there is no additional gain to setting things in QM.slice, as changes to a single qm.container file would work the same way.

I agree that is is somewhat weird to have qm.service directly under the root compared to e.g. system.slice, but I think the disadvantages of having qm.slice outweigh that. I mean, we, the qm developers themselves shoot themselves in the foot by it, not realizing that CPUWeight was completely broken for more than a year, and requiring deep debugging to even figure out what caused this.

Can you imaging the docs trying to explain which options should go where to make them work? You'd have to have a PhD in cgroups to understand it.  Whereas otherwise it will just say "change all options (cgroups or not) in qm.container".

> so, If IIUC correctly, you are recommending drop QM.slice. (MHO, there is a good reason to maintain it)

You didn't say what the good reason is? I mean, it is not "you can easily modify the cgroups of all processes running within the QM container environment", because that is already possible by just editing the qm.container file. Having both the .slice and the .container file just gives you *two* places to set the same options, except 10% of the options silently don't work if you accidentally change the wrong file.

Yarboa:
> All the current options what we change, including he current value of CPUWeight (which doesn't work) are in the qm.container file, and any changes to this file will affect _all_ qm processes (at least if we drop qm.slice). The later means there is no additional gain to setting things in QM.slice, as changes to a single qm.container file would work the same way.
> 
> I agree that is is somewhat weird to have qm.service directly under the root compared to e.g. system.slice, but I think the disadvantages of having qm.slice outweigh that. I mean, we, the qm developers shoot ourselves in the foot by it, not realizing that CPUWeight was completely broken for more than a year, and requiring deep debugging to even figure out what caused this.
> 
> Can you imaging the docs trying to explain which options should go where to make them work? You'd have to have a PhD in cgroups to understand it. Whereas otherwise it will just say "change all options (cgroups or not) in qm.container".
> 
> > so, If IIUC correctly, you are recommending drop QM.slice. (MHO, there is a good reason to maintain it)
> 
> You didn't say what the good reason is? I mean, it is not "you can easily modify the cgroups of all processes running within the QM container environment", because that is already possible by just editing the qm.container file. Having both the .slice and the .container file just gives you _two_ places to set the same options, except 10% of the options silently don't work if you accidentally change the wrong file.

OK LGTM, my side,
- Let me check if there are some tests related to QM mentions QM.slice, 
- what about docs? shall we create different issue for that?

alexlarsson:
@Yarboa I updated the docs here

Yarboa (approved):
LGTM

rhatdan:
If I want to change the CPUWeight for all processes within the QM, does this need to change?

Yarboa:
> @rhatdan What do you mean exactly? This value is the default for all processes, and if you want to change it for all the processes in qm you need to override this value (by setting CPUWeight in a qm.container.d drop-in file). This is similar to how all other config options in qm works, like e.g. adding a volume or setting the max memory use.

As per the man systemd.resource-control
If there are slices create under QM, Delegate is needed, isnt it?

alexlarsson:
@Yarboa Delegate is needed to allow podman to create sub-cgroups, yes. However, strictly speaking it doesn't have to be in the container file, because quadlet always adds it to the generated service file. Honestly, we should probably remove it from qm.container.

rhatdan:
I am thinking 
systemctl set-property <unit_name> CPUWeight=<value> 

Basically is there a way for a process monitor to modify the CPUWeight of the entire QM.  I thought we were doing this in the demos with

systemctl set-property QM.slice CPUWeight=X

PR #750, opened 2025-03-17
Update NETWORK.md missing documentation.

nsednev:
I've changed `podman run -it fedora /bin/bash` to `podman run -it --network=private fedora /bin/bash` that was missing.

## Summary by Sourcery

Documentation:
- Correct the example command in the NETWORK.md tutorial to include the `--network=private` option when running a container with its own isolated network namespace.


rhatdan:
The default is usually --network=private, so this probably does not change much ,but does it make it clearer.
LGTM

PR #751, opened 2025-03-17
Change CPUWeight default to "idle"

alexlarsson:
This default means that the QM partition only gets CPU if nothing is wants to run on the rest of the system. Essentially it gets SCHED_IDLE behaviour.

## Summary by Sourcery

Enhancements:
- Change the default CPUWeight setting to 'idle' for the QM partition, ensuring it only receives CPU resources when the rest of the system is idle.

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -11,7 +11,7 @@ WantedBy=default.target
 # ManagedOOMSwap=auto|kill: Specifies how systemd-oomd.service will act on qm.
 # QM cgroup, pass directly to systemd and handled by it,
 # please refer to `man systemd.resource-control` for details.
-CPUWeight=50
+CPUWeight=idle
 Delegate=true
 IOWeight=50
 ManagedOOMSwap=kill


rhatdan:
Does the QM still perform well in this situation?

alexlarsson:
@rhatdan In an idle system there should be no difference. But for sure it will depend on the ASIL load. At the end of the day though, this is really only a change in the default behaviour. If you need/want a different behavior you can set that up. For example, and alternative setup is to use CPU pinning and give QM full access to some subset of CPUs.

alexlarsson:
This seems to fail:

> 14:50:20                 cmd: /bin/bash ./check_bluechi_is_ok.sh
14:50:20         environment
14:50:20                 out: [ INFO  ] FAIL: check_bluechi_is_ok: host bluechi-agent is not connected to controller.
14:50:20                 out:  Mar 17 14:50:16 ip-172-31-28-176.us-east-2.compute.internal bluechi-controller[24312]: Node 'qm.localrootfs' disconnected
14:50:20                 out: Mar 17 14:50:17 ip-172-31-28-176.us-east-2.compute.internal bluechi-controller[24312]: Registered managed node from fd 10 as 'qm.localrootfs'
14:50:20                 out: Shared connection to 52.14.158.158 closed.
14:50:20         Command returned '1' (failure).

I wonder why?

dougsland:
> This seems to fail:
> 
> > 14:50:20                 cmd: /bin/bash ./check_bluechi_is_ok.sh
> > 14:50:20         environment
> > 14:50:20                 out: [ INFO  ] FAIL: check_bluechi_is_ok: host bluechi-agent is not connected to controller.
> > 14:50:20                 out:  Mar 17 14:50:16 ip-172-31-28-176.us-east-2.compute.internal bluechi-controller[24312]: Node 'qm.localrootfs' disconnected
> > 14:50:20                 out: Mar 17 14:50:17 ip-172-31-28-176.us-east-2.compute.internal bluechi-controller[24312]: Registered managed node from fd 10 as 'qm.localrootfs'
> > 14:50:20                 out: Shared connection to 52.14.158.158 closed.
> > 14:50:20         Command returned '1' (failure).
> 
> I wonder why?

Lets retrigger it.

dougsland:
> /packit retest-failed

@engelmi I will add you as write access.

Yarboa:
> > /packit retest-failed
> 
> @engelmi I will add you as write access.

The connection from qm bluechi-agent
 disconnected
Node 'qm.localrootfs' disconnected

engelmi:
> > > /packit retest-failed
> > 
> > 
> > @engelmi I will add you as write access.
> 
> The connection from qm bluechi-agent disconnected Node 'qm.localrootfs' disconnected

Question would be why. This PR doesn't change anything related to it. 
Is the CI pipeline using a version of QM from COPR or one built from the respective PR? @Yarboa My guess would be it uses a COPR version so we get failures there delayed - this PR seems more related to the failure: https://github.com/containers/qm/pull/749

dougsland:
> > > > /packit retest-failed
> > > 
> > > 
> > > @engelmi I will add you as write access.
> > 
> > 
> > The connection from qm bluechi-agent disconnected Node 'qm.localrootfs' disconnected
> 
> Question would be why. This PR doesn't change anything related to it. Is the CI pipeline using a version of QM from COPR or one built from the respective PR? @Yarboa My guess would be it uses a COPR version so we get failures there delayed - this PR seems more related to the failure: #749

The #749 seems to have passed in CI/CD just fine.

engelmi:
> The #749 seems to have passed in CI/CD just fine.

Thats why I was wondering if the CI uses a COPR version of QM, which would be a passing one at the time of the previous PR.

dougsland:
> > The #749 seems to have passed in CI/CD just fine.
> 
> Thats why I was wondering if the CI uses a COPR version of QM, which would be a passing one at the time of the previous PR.

@engelmi @Yarboa I believe it's CI/CD load, retriggering twice made one fail at time pass. Looks like we have green light.

Yarboa:
> > The #749 seems to have passed in CI/CD just fine.
> 
> Thats why I was wondering if the CI uses a COPR version of QM, which would be a passing one at the time of the previous PR.

please refer 
info_message 'Installing qm using packit/containers-qm-751 repo'
under tmt-verbose.log
other why do we need CI?

dougsland:
@alexlarsson I am starting wondering if the issues we saw is related to SCHED_IDLE where it only get priority if the CPU is idle. In this case bluechi is a container inside QM. Which brings to think if SCHED_OTHER would be a best approach?

alexlarsson:
@dougsland I would prefer to have the more FFI-safe and customer-expected default over wild guesses about the cause of this. Idle scheduling shouldn't affect this unless the non-qm side is using 100% cpu, which seems unlikely in this test. And even then, it should only delay the qm bluechi-agent, not disconnect it.

I think we need to try to reproduce locally to figure out what is actually happening.

PR #753, opened 2025-03-17
[TEST] set other as CPUWeight

dougsland:
## Summary by Sourcery

Tests:
- Adds a test case for setting CPU weight.

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -11,7 +11,7 @@ WantedBy=default.target
 # ManagedOOMSwap=auto|kill: Specifies how systemd-oomd.service will act on qm.
 # QM cgroup, pass directly to systemd and handled by it,
 # please refer to `man systemd.resource-control` for details.
-CPUWeight=idle
+CPUWeight=other
 Delegate=true
 IOWeight=50
 ManagedOOMSwap=kill


dougsland:
@alexlarsson @Yarboa @rhatdan @engelmi CI/CD passed in the first try.

Yarboa:
> @alexlarsson @Yarboa @rhatdan @engelmi CI/CD passed in the first try.

I have to do some reading here, 
but @alexlarsson and

Yarboa:
> @alexlarsson @Yarboa @rhatdan @engelmi CI/CD passed in the first try.

BUt from man systemd.resource-control

If set to the special string "idle", mark the cgroup for "idle scheduling", which means that it
               will get CPU resources only when there are no processes not marked in this way to execute in
               this cgroup or its siblings.
              
For the other, i do not identify it in that man, is it a valid parameter? do you have a link backing up this config?

pypingou:
This is basically undoing the investigation we did over the past few days. We do want QM to default to the `idle` scheduler by default. It can be changed back to `other` sure, but that would be an user's choice based on the workload they want to run on the system. To demonstrate the FFI that we want to show, we should default to `idle`.

Yarboa:
I assume other means default of CPUWeight=100
since it is not "understood"

dustinblack:
I'm a little surprised this passed CI since I don't know `other` to be a valid value. My suspicion is that this has the same effect as being unset or set to 100.

engelmi:
> I'm a little surprised this passed CI since I don't know `other` to be a valid value. My suspicion is that this has the same effect as being unset or set to 100.

It seems systemd doesn't know about it and is skipping it
```bash
fedora systemd[1]: /etc/systemd/system/test.service:6: Invalid CPU weight, ignoring: other
```

dougsland:
> This is basically undoing the investigation we did over the past few days. We do want QM to default to the `idle` scheduler > by default. It can be changed back to `other` sure, but that would be an user's choice based on the workload they want to
> run on the system. To demonstrate the FFI that we want to show, we should default to `idle`.

Ok, lets keep this way @pypingou

Issue #717, opened 2025-02-06
Please document how to change the network using quadlet's.


dougsland:
It's different scenario but good to have.

See-Also: https://github.com/containers/qm/pull/713

nsednev:
After our discassion with @dougsland , adding how this should be implemented as quadlet like realtime.

It should be implemented using [realtime](https://github.com/containers/qm/?tab=readme-ov-file#realtime) but with Network session.

should be Network=host or Network=private 
restart qm and that's it
podman restart qm
ip addr show
should help to show the difference
just expose 2 systemcalls that are blocked by default
Should be sched_setaffinity and sched_getaffinity
In realtime, there is no blockage.

nsednev:
[MR](https://github.com/containers/qm/pull/776) has been created for this ticket.

nsednev:
All suggestions were answered and resolved.
Waiting for final review so we could push an MR and close the ticket.

nsednev:
Closing this ticket as the [MR](https://github.com/containers/qm/pull/776) has been merged.

PR #776, opened 2025-03-27
Added network README.md for network options using quadlets.

nsednev:
Here is an example of running a network-test container using quadlets for both --network=host and --network=private.
This is a README file as was requested in https://github.com/containers/qm/issues/717 .

## Summary by Sourcery

Documentation:
- Create a README.md file demonstrating how to configure container networks using quadlets, showing examples for host and private network modes

### DIFF
--- a/docs/quadlet-examples/network/README.md
+++ b/docs/quadlet-examples/network/README.md
@@ -1,0 +1,35 @@
+# Here is an example of running a network-test container using quadlets for both --network=host and --network=private
+
+You should place this file either in /usr/share/containers/systemd/ or /etc/containers/systemd/
+
+```console
+/usr/share/containers/systemd/
+/etc/containers/systemd/
+```
+
+For rootless users:
+
+```console
+$HOME/.config/containers/systemd/
+
+```
+
+Host Network
+
+```console
+# network-test.container
+[Container]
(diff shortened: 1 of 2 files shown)


Yarboa (commented):
Does it related to ? 
https://github.com/containers/qm/blob/main/docs/tutorials/NETWORK.md

Why not adding it at then of that one? 
Is there a reference in Network.md to to README.md?

nsednev:
> Does it related to ? https://github.com/containers/qm/blob/main/docs/tutorials/NETWORK.md
> 
Yes it is related, it provides basic example running container with either --network=host or --network=private modes using quadlets.
> Why not adding it at then of that one? Is there a reference in Network.md to to README.md?

- The reason that it has been separated and placed to `docs/quadlet-examples` is because we already have it specifically for that purpose and after we discussed this with @dougsland. I thinks that quadlet examples should be separated from the main network document, in which we may add later on quadlets configuration section, which will explain in details how to work with quadlets, but the quadlet examples ready to be used should be separated.

- The reference from NETWORK.md to this README.md has been added.

Yarboa (approved):
LGTM

dougsland (approved):

PR #777, opened 2025-03-27
docs/devel: cpu management

iamianmullins:
## Summary by Sourcery

Add documentation for managing CPU usage of the qm.service using systemd attributes like CPUWeight and CPUQuota

New Features:
- Introduce methods to control CPU resource allocation for qm.service using systemd attributes

Documentation:
- Add detailed instructions for managing CPU usage of qm.service through systemd configuration
- Provide step-by-step guidance on setting CPUWeight and CPUQuota for service prioritization and CPU limit

### DIFF
--- a/docs/devel/README.md
+++ b/docs/devel/README.md
@@ -13,6 +13,7 @@
   - [Listing QM service](#listing-qm-service)
   - [List QM container via podman](#list-qm-container-via-podman)
   - [Extend QM quadlet managed by podman](#extend-qm-quadlet-managed-by-podman)
+  - [Managing CPU usage](#managing-cpu-usage)
   - [Connecting to QM container via podman](#connecting-to-qm-container-via-podman)
   - [SSH guest CentOS Automotive Stream Distro](#ssh-guest-centos-automotive-stream-distro)
   - [Check if HOST and Container are using different network namespace](#check-if-host-and-container-are-using-different-network-namespace)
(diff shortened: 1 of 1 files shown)


docs/devel/README.md
```diff
@@ -372,3 +372,82 @@ bash-5.1#
...
+Using the steps below, it's possible to manage CPU usage of the `qm.service` by modifying service attributes and utilizing drop-in files.
+
+#### Setting the CPUWeight attribute
+
+Modifying the `CPUWeight` attribute affects the priority the of `qm.service`. A higher value prioritizes the service, while a lower value deprioritizes it.
+
+Inspect the current CPUWeight value:
+
+```bash
+systemctl show  qm.service | grep CPUWeight
```
Yarboa, line 387: Please use that one instead
systemctl show -p CPUWeight  qm.service

docs/devel/README.md
```diff
@@ -372,3 +372,82 @@ bash-5.1#
...
+```
+
+#### Limiting CPUQuota
+
+It's also possible to limit the percentage of the CPU allocated to the `qm.service` by defining `CPUQuota`. The percentage specifies how much CPU time the unit shall get at maximum, relative to the total CPU time available on one CPU.
+
+Inspect the current `CPUQuota` value:
+
+```bash
+systemctl show  qm.service | grep CPUQuota
```
Yarboa, line 403: same as above

docs/devel/README.md
```diff
@@ -372,3 +372,82 @@ bash-5.1#
...
+Set the `CPUQuota` value of `qm.service` on the host using:
+
+```bash
+systemctl set-property qm.service CPUQuota=50%
+```
+
+Verify the `CPUQuota` drop in file has been created using the command below.
+
+```bash
+systemctl show  qm.service | grep CPUQuota
```
Yarboa, line 415: same as above

docs/devel/README.md
```diff
@@ -372,3 +372,82 @@ bash-5.1#
...
+CPUQuotaPeriodUSec=infinity
+DropInPaths=/usr/lib/systemd/system/service.d/10-timeout-abort.conf /etc/systemd/system.control/qm.service.d/50-CPUQuota.conf
+```
+
+To test maxing out CPU usage and then inspect using the `top` command, follow these steps:
+
+- From within the qm container, execute this command to stress the CPU:
+
+```bash
+dd if=/dev/zero of=/dev/null
```
Yarboa, line 431: Consider using that one instead
podman exec qm dd if=/dev/zero of=/dev/null

docs/devel/README.md
```diff
@@ -372,3 +372,82 @@ bash-5.1#
 ```
 
 As you can see above, the error occurs because the Quadlet is attempting to use an unsupported key from the Service section in the Container group. Removing the unsupported key ```Command``` from ```ros2-rolling.container``` and then reloading or restarting the service should resolve the issue.
+
+### Managing CPU usage
```
Yarboa, line 376: MHO, section should move to line L#216
Under that section. 
### Extend QM quadlet managed by podman

docs/devel/README.md
iamianmullins: Thanks for the suggestion, I've added the `timeout 30` also, otherwise `dd` continues to run

PR #784, opened 2025-04-04
tools: just examples of cpu/mem usage

dougsland:
Example of collecting cpu/mem using cgroups v2

## Summary by Sourcery

New Features:
- Create example scripts to demonstrate CPU and memory usage collection methods using cgroups v2

### DIFF
--- a/tools/getcpu-usage
+++ b/tools/getcpu-usage
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+# flake8: noqa: E501
+
+import os
+import time
+
+
+def count_cpus(cpuset_str):
+    total = 0
+    for part in cpuset_str.split(','):
+        if '-' in part:
+            start, end = map(int, part.split('-'))
+            total += end - start + 1
+        else:
+            total += 1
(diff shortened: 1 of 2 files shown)


Yarboa (commented):
Please Add README explaining this tools

dougsland:
> Please Add README explaining this tools

These tools, you just run to get cpu consume and memory consume. Should be the same systemd-cgtool

Yarboa:
> > Please Add README explaining this tools
> 
> These tools, you just run to get cpu consume and memory consume. Should be the same systemd-cgtool

So why is it needed in that repo? 
Should be explained somewhere? who is the audience it is targeting?

dougsland:
> > > Please Add README explaining this tools
> > 
> > 
> > These tools, you just run to get cpu consume and memory consume. Should be the same systemd-cgtool
> 
> So why is it needed in that repo? Should be explained somewhere? who is the audience it is targeting?

To help identify memory and cpu in the code and quick access to users. However, let me close this one.

PR #788, opened 2025-04-06
Updated qm manpage to fix install instructions.

nsednev:
This MR fixes QM installation instructions in /qm/qm.8.md file.


nsednev:
Closing this MR as no need to modify current documentation as it works properly as it is now.

PR #790, opened 2025-04-06
subpackages: adjust path and remove qm_rootfs

dougsland:
## Summary by Sourcery

Adjust the configuration paths and remove references to qm_rootfs in the windowmanager RPM spec file

Enhancements:
- Replace hardcoded rootfs path with a more flexible sysconfdir configuration
- Update version macro to use dynamic versioning

Chores:
- Refactor RPM spec file to use more standard RPM macro practices


Yarboa (approved):
LGTM, one question here, 
I assume AddDevice should be added to the container service file similar to QM quadlet file 
Did you check it?

PR #811, opened 2025-04-16
qm.container: drop sys_boot capability

aesteve-rh:
A better solution would be to rely on namespaces,
but in the meantime, this helps clearing a
safety concern from assessors.

## Summary by Sourcery

Enhancements:
- Reduce container privileges by dropping the sys_boot capability


dougsland:
LGTM

Issue #826, opened 2025-05-14
improve /tests/ffi/deny_set_scheduler according to #824


pengshanyu:
According to this PR:  sched_setscheduler to allow other, batch idle #824 
The test /tests/ffi/deny_set_scheduler needs to be improved accordingly.

pengshanyu:
Submitted PR:  
https://github.com/containers/qm/pull/832
https://github.com/containers/qm/pull/833

PR #824, opened 2025-05-08
sched_setscheduler to allow other, batch idle

dougsland:
The current seccomp changes completely disallow
calling sched_setscheduler, but we can safely
allow calling it with policy==SCHED_OTHER/BATCH/IDLE, as really the only problem is the various real-time classes.

The profile argument is the second (id 1) and the values for the classes are OTHER=0, BATCH==3, IDLE==5

Resolves: #702
See-Also: https://github.com/containers/qm/pull/818

PR #833, opened 2025-05-20
improve deny_set_scheduler (part 2)

pengshanyu:
Improved test qm/tests/ffi/deny_set_scheduler according to PR https://github.com/containers/qm/pull/824

I divided this PR into 2 parts. This is the second part of the PR, which modifies qm/tests/ffi/deny_set_scheduler/test.sh

After PR #832  merged, this PR should make the gate test pass.

Resolves: https://github.com/containers/qm/issues/826

## Summary by Sourcery

Refine the deny_set_scheduler test script to iterate over all scheduling policies, assign policy-specific expected outcomes, and aggregate pass/fail results.

Enhancements:
- Loop through SCHED_OTHER, SCHED_BATCH, SCHED_IDLE, SCHED_FIFO, and SCHED_RR policies instead of testing a single case
- Determine expected success or failure dynamically based on policy and compare against actual output
- Accumulate failures across policies and return a consolidated test result

Tests:
- Update test.sh to copy the test binary into the qm container before execution
- Track the number of failed policy checks and exit with an appropriate status

### DIFF
--- a/tests/ffi/deny_set_scheduler/test.sh
+++ b/tests/ffi/deny_set_scheduler/test.sh
@@ -3,19 +3,44 @@
 # shellcheck source=tests/ffi/common/prepare.sh
 . ../common/prepare.sh
 
-expected_result="Failed to set scheduler: Function not implemented"
+policy_arr=(SCHED_OTHER SCHED_BATCH SCHED_IDLE SCHED_FIFO SCHED_RR)
+num_of_failed_policy=0
 
 trap disk_cleanup EXIT
 prepare_test
 reload_config
 
 running_container_in_qm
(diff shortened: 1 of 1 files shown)


Yarboa:
Advise,
Rebase this PR with dependent one, then  you could now if test really works.
When you git push force, it will contain both PRs
When the dependent PR is merged, this PR will reamin only the diff from main

pengshanyu:
Hi @Yarboa , thank you so much for the advice. Only when the modifications of qm/tests/e2e/tools/FFI/deny_set_scheduler/execute_set_scheduler.c reaches ffi-tools, then we can see this test really works. So I manually tested the modification before submitting the PR.

PR #843, opened 2025-06-16
qm.fc: change context for ipc programs

dougsland:
Today we have /var/run/ipc, changing to /run/ipc.


dougsland:
Closing, Dan sent already.

PR #848, opened 2025-06-18
selinux: add additional selinux for ipc

### LINKED ISSUE #840: SELinux policy for QM prevents communications with IPC sockets

```
```

dougsland:
The test is qm-to-qm ipc

resolve #840 

## Summary by Sourcery

New Features:
- Add SELinux policy rules to allow IPC (shared memory, semaphores, message queues) between QEMU instances

### DIFF
--- a/qm.te
+++ b/qm.te
@@ -30,7 +30,19 @@ unconfined_domain(ipc_t)
 
 qm_domain_template(qm)
 
+#########################################
+# IPC between QM containers
+#########################################
 
+require {
+    type qm_file_t;
+    type qm_container_ipc_t;
+    class sock_file write;
+}
+
+#============= qm_container_ipc_t ==============
+allow qm_container_ipc_t qm_file_t:sock_file write;
+allow qm_container_ipc_t qm_t:unix_stream_socket connectto;
 
 #########################################
 #


dougsland:
Looks like it's working with the new selinux rules:

Client side:
```
# podman logs systemd-ipc_client
Connecting to: /run/ipc/ipc_server.socket
```

Server side:
```
# podman logs systemd-ipc_server
Current time is: 15:38:51
Current time is: 15:38:52
Current time is: 15:38:53
Current time is: 15:38:54
Current time is: 15:38:55
Current time is: 15:38:56
Current time is: 15:38:57
Current time is: 15:38:58
Current time is: 15:38:59
Current time is: 15:39:00
Current time is: 15:39:01
Current time is: 15:39:02
Current time is: 15:39:03
Current time is: 15:39:04
Current time is: 15:39:05
Current time is: 15:39:06
Current time is: 15:39:07
Current time is: 15:39:08
Current time is: 15:39:09
Current time is: 15:39:10
```

qm.te
```diff
@@ -30,7 +30,19 @@ unconfined_domain(ipc_t)
...
+
+require {
+    type qm_file_t;
+    type qm_container_ipc_t;
+    class sock_file write;
+}
 
+#============= qm_container_ipc_t ==============
+allow qm_container_ipc_t qm_file_t:sock_file write;
+allow qm_container_ipc_t qm_t:unix_stream_socket connectto;
```
engelmi, line 45: I think we can replace this by the [stream_connect_pattern](https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/ipc_patterns.spt#L4-L8) from the fedora/selinux-policy, adding also `search_dir_perms` on the ipc directory:
```suggestion
stream_connect_pattern(qm_container_ipc_t, ipc_var_run_t, qm_file_t, qm_t);
```

qm.te
```diff
@@ -30,7 +30,19 @@ unconfined_domain(ipc_t)
...
 
+#########################################
+# IPC between QM containers
+#########################################
+
+require {
+    type qm_file_t;
+    type qm_container_ipc_t;
+    class sock_file write;
+}
```
engelmi, line 41: This `require` block can be removed, I think. 
Both types are defined in the qm policy - `qm_file_t` ([here](https://github.com/containers/qm/blob/main/qm.if#L52)) and `qm_container_ipc_t` ([here](https://github.com/containers/qm/blob/main/qm.if#L329)). And the `sock_file` class is being used further down by the stream_connect_pattern of the bluechi part of the policy.

qm.te
dougsland: macro seems much cleaner, I will give a test for double check. Thanks @engelmi

PR #857, opened 2025-06-30
qm.if: add ipc mounton rule

dougsland:
allow $1_container_ipc_t ipc_var_run_t:dir mounton;


rhatdan:
This is fine, but what is qm_container_ipc_t mounting?  Is this a podman command running?

rhatdan:
Yes all /var/run labeling needs to be moved to /run.

PR #861, opened 2025-07-07
qm.te: add allow for IPC (qm to qm)

### LINKED ISSUE #840: SELinux policy for QM prevents communications with IPC sockets

dougsland:
## Summary by Sourcery
fix #840 
ipc communication inside QM
as documented in 
[qm-to-app](https://github.com/containers/qm/blob/c4f1ed9cf0ba10acd2ec38ff84264d68e98da0c2/docs/docs/ipc.md#example-qm-to-qm-app)

New Features:
- Permit IPC communication between qm processes in the SELinux policy

### DIFF
--- a/qm.if
+++ b/qm.if
@@ -414,6 +414,16 @@ template(`qm_domain_template',`
 	allow $1_t $1_container_wayland_t:dbus send_msg;
 	dev_read_sysfs($1_container_wayland_t)
 
+        #########################################
+        #
+        #   QM to QM IPC Communication
+        #
+        # Allow rw access to sock files labeled qm_file_t
+        allow $1_container_t $1_file_t:sock_file { read write };
+        # Allow container to connect to unix_stream_socket of qm_t processes
+        allow $1_container_t $1_t:unix_stream_socket connectto;
+
+
 	allow getty_t $1_file_type:chr_file rw_chr_file_perms;
 
 	read_files_pattern(systemd_hostnamed_t, $1_file_t, $1_file_t)


qm.if
```diff
@@ -405,6 +405,15 @@ template(`qm_domain_template',`
...
+        #
+        #   QM to QM IPC Communication
+        #
+        # Allow rw access to sock files labeled qm_file_t
+        allow $1_container_t $1_file_t:sock_file { read write };
+        # Allow container to connect to unix_stream_socket of qm_t processes
+        allow $1_container_t $1_t:unix_stream_socket connectto;
+
 	allow $1_container_wayland_t $1_file_t:chr_file map;
 	manage_dirs_pattern($1_container_wayland_t, $1_file_t, $1_file_t)
```
Yarboa, line 418: try to maintain the weyland part in one paragraph
You just cut it

Yarboa:
@dougsland verified manually with 
dnf copr enable  packit/containers-qm-861 epel-9-x86_64

rhatdan:
So rather then having communications between a couple of qm_container_t processes, you want to create a qm_t process running outside of a container and then allow all of the containers to communicate with it?

This seems like an expansion of the requirements, we talked about. I am fine with allowing this, but you need to understand what you are allowing.

Yarboa:
> So rather then having communications between a couple of qm_container_t processes, you want to create a qm_t process running outside of a container and then allow all of the containers to communicate with it?
> 
> This seems like an expansion of the requirements, we talked about. I am fine with allowing this, but you need to understand what you are allowing.

no, not for that use case, 
Contianers and sockets are all in qm here

dougsland:
> > So rather then having communications between a couple of qm_container_t processes, you want to create a qm_t process running outside of a container and then allow all of the containers to communicate with it?
> > This seems like an expansion of the requirements, we talked about. I am fine with allowing this, but you need to understand what you are allowing.
> 
> Ok, so based on your comment no need selinux updates here I have added the following to unit files for client and server
> 
> SecurityLabelLevel=s0:c1,c2 SecurityLabelFileType=qm_container_file_t
> 

For me still not flying. I am trying to figure out.

> @dougsland please update docs here https://github.com/containers/qm/blob/6dd6ec4fa488c577d2c7ea1325324bd3c1746927/docs/docs/ipc.md#etcqmcontainerssystemdipc_clientcontainer
> 
> Need to add to quadlets SecurityLabelFileType=qm_container_file_t

I will update for sure but need to make 100% working first.

aesteve-rh:
Communicating two `qm_container_ipc_t` labelled containers can be achieved by just adding:
```
allow qm_container_ipc_t qm_file_t:sock_file write;
```
I think we need to decide what scenario we do want to support, maybe more than one option is possible.

Yarboa:
closing that one due to #877

PR #871, opened 2025-07-12
setup: add support to libkrun

dougsland:
setup is used in CI/CD, good to have such support.

## Summary by Sourcery

Enhancements:
- Include installation and configuration steps for libkrun in the setup script


Yarboa (commented):
Please update the libkrun test
In case this one is in, the tests should check if rpms already installed,
In case not it should continue with how it is written

setup
```diff
@@ -369,6 +377,10 @@ case "$1" in
     else
       /usr/libexec/podman/quadlet /run/systemd/generator/
     fi
+
+    if [ "$LIBKRUN_SKIP" == "N" ]; then
```
dougsland, line 381: == "Y" ?

Yarboa: Moved back to your initial suggestion

Issue #873, opened 2025-07-14
Reduce message queues capacity


pridolfi:
Running interference tests we discovered an issue with message queues. It's possible to exhaust memory assigned to mq from QM, and make mq creation from the root layer impossible.

Suggestion: reduce number of max mqueues to `4`.

```
[Container]
Sysctl=fs.mqueue.queues_max=4
```

PR #874, opened 2025-07-14
qm.container: Limit max mqueues to 4

aesteve-rh:
Limit the max number of message queues to 4.
Create a test to verify this setting.

Fixes: https://github.com/containers/qm/issues/873

## Summary by Sourcery

Enforce a maximum of 4 POSIX message queues in the QM container and add a comprehensive FFI test to verify the setting and its enforcement

New Features:
- Set fs.mqueue.queues_max sysctl in the QM container to limit message queues to 4

Documentation:
- Add a README describing the purpose and steps of the message queue limit test

Tests:
- Add a C test program and shell script to validate the sysctl value and enforce the 4-queue limit
- Include a Makefile and FMF metadata for building and running the message queue limit test

### DIFF
--- a/qm.container
+++ b/qm.container
@@ -91,6 +91,17 @@ Mount=type=tmpfs,destination=/dev/shm
 #
 Rootfs=${ROOTFS}
 
+# The value for fs.mqueue.queues_max determines the maximum number of POSIX message queues
+# that can be created system-wide within the container. This value is set based on available
+# system resources and the expected usage pattern. The calculation is approximately:
+#   queues_max ≈ (system-wide RLIMIT_MSGQUEUE) / (msg_max * msgsize_max) / 2
+# where:
+#   - RLIMIT_MSGQUEUE is the total bytes allowed for all message queues (ulimit -q)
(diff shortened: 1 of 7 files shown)


aesteve-rh:
The issue has the context, but basically the reasoning is to avoid the qm exhausting the message queues for the root layer.

Withe default `msgsize_max` of `8192` and an `ulimit` of `819200`, we can only fit up to 9 buffers simultaneously. By limiting the QM to 4 we ensure that more than half of the available memory is reserved for the ASIL layer.

@dougsland @Yarboa I think this has been green-lighted already. So feel free to review when you have time.

imcleod:
@aesteve-rh Is there a way to add something to this test that confirms that within the ASIL/root partition, we have a different limit and can still create message queues after QM has exhausted its limit? I don't see anything like that in the MR right now.

I ask because the documentation on this setting is slightly ambiguous. The top level docs on the /proc interface suggest it is system-wide, however the man page on IPC namespaces suggests that each namespace gets its own set of values here, though they may be inherited from the parent namespace at creation time.

Yarboa (commented):
LGTM

PR #875, opened 2025-07-14
Limit the number of SELinux calls qm containers can do

rhatdan:


rhatdan:
BTW these are the same access allowed for container_t now.

PR #877, opened 2025-07-16
Update qm-2-qm ipc rules

### LINKED ISSUE #840: SELinux policy for QM prevents communications with IPC sockets

Yarboa:
This issue Resolve #840

Containers - Container IPC inside qm does not need systemd socket
For qm-2-qm no qm_t systemd socket mapping qm_file_t -> qm_container_file_t

## Summary by Sourcery

Refine QM SELinux policy by removing systemd socket dependency for container IPC and adding distinct file labeling for container files to enforce updated qm-2-qm communication rules.

Enhancements:
- Remove unnecessary systemd socket requirement for inter-container IPC within qm
- Introduce qm_container_file_t file context to replace qm_file_t for container files
- Update qm SELinux policy and interface definitions to enforce new QM-to-QM IPC rules

### DIFF
--- a/docs/docs/ipc.md
+++ b/docs/docs/ipc.md
@@ -30,7 +30,7 @@ while maintaining SELinux isolation.
 On the other hand, **IPC between QM services** (e.g., two services or containers within the same QM domain)
 can occur as well. Since these components share the same SELinux type and context, they are allowed to
 communicate using standard Unix domain sockets located in /run. This approach simplifies internal QM
-communication rely on containers ipc without compromising the system's overall security posture. Such communication can be
+communication relies on container IPC without compromising the system's overall security posture. Such communication can be
 orchestrated also using container orchestration patterns like **.pod (Podman pod definitions)** or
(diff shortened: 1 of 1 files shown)


docs/docs/ipc.md
```diff
@@ -107,8 +104,7 @@ Volume=/run/ipc/:/run/ipc/
 ```console
 [Unit]
 Description=Demo client service container
-Requires=ipc_server.socket
-After=ipc_server.socket
+
```
aesteve-rh, line 107: Maybe have?
```suggestion
Requires=ipc_server.service
After=ipc_server.service

```

Having this dependency would allow it to mount the socket created by the server directly with:
```
Volume=/run/ipc.socket:/run/ipc.socket
```

Yarboa:
> * om the old ipc_server.socket un

Good point
We have another option to connect through systemd sockets, but it is future work
Can you suggest how to mention it under the ipc.md?

I will create an issue on future work related to your comment
The missing part is one to many clients communication, which is not part of the guide for now

PR #888, opened 2025-07-21
Remove application specific settings

shetze:
The QM SELinux policy is not the right place to assign application-specific rights. If there is a systematic problem with the QM policy, it should be solved independently of the specific application with which the problem was first identified.

However, if the problem should actually be solved upstream in the application, it makes sense to document in the qm README what constitutes good practice for adapting to the QM partition concept.

## Summary by Sourcery

Remove application-specific rights from the QM SELinux policy and document recommended adaptation practices in the README

Enhancements:
- Strip out application-specific settings from qm.te and qm.if policy modules

Documentation:
- Add guidance to the qm README for adapting applications to the QM partition concept


Yarboa:
@hellocthulhu 
I think we have to coordinate that PR with Automotive-sig docs
WDYT?

hellocthulhu:
hi @Yarboa There are a few places where we mention SELinux in the docs, but I am not sure if the existing content needs to be updated or if new content needs to be created to resolve this issue in Auto SIG docs. Here are some resources. Take a look & let me know what you think

- [Enhancing security for containers with SELinux in the QM partitio](https://sigs.centos.org/automotive/features-and-concepts/con_mixed-criticality/#enhancing-security-for-containers-with-selinux-in-the-qm-partition)

Yarboa:
> I created these issues in Automotive SIG docs repo. Please add details to the issues for context. [Issue 80](https://gitlab.com/CentOS/automotive/sig-docs/-/issues/80) - Deploying sample applications in containers deserves a refresher [Issue 81](https://gitlab.com/CentOS/automotive/sig-docs/-/issues/81) - Configuring communication between containers in QM and root partitions deserves refresher
> 
> If new topics are required for SELinux-related work, please open a GitLab issue or speak with Alex McLeod.

+2 @shetze,
This sample. created long before QM env was introduced, and evolved.
I see here few points.
1. the vsomeip should be converted to to work with latest conventions related to -v /run/ipc:/run/ipc dir.
2. In case nee selinux rules needed, some demo should be explained
Which files to update, how to create app-selinux.pp rules how to install and how to remove it.
Not mentioned at all in docs, vsomeip app and how to modify selinux rules could be explained in docs. 

Note: the app for creating and modifying is good, but i tend to expose the audience for how to create selinux rules for 
3. related to vsomeip app and security holes, further discussion needed

> One maybe last questions with regards to the vsomeip example is, if it may be allowed to communicate from QM to ASIL. In the current example implementation, both sides want to log via dlt. As I see it, dlt itself carries safety requirements if logging for a safety application. So we either end up with two instances of dlt, one on each side, or we allow the QM applications to write log messages into the one ASIL instance dlt socket. My guess is we end up with two completely independent dlt instances. But this should be decided by the safety experts dealing with AUTOSAR Adaptive.

alexlarsson:
I haven't looked at this discussion in detail, but here are some thoughs:

The vsomeip policy is some ancient work that was done to create examples (and to test things out) for how we can allow communication between the qm partition and the outside. A vsomeip implementation where all clients and routers are in the qm partition would need no such policy. I agree that putting this policy in the qm package is incorrect, because it is very specific to a particular example package (vsomeip) that isn't ASIL-B certified, or even actually in our product at all. Additionally it is a poor way to demonstrate how to do this, because a user can not use the same approach (i.e. couldn't be added to the qm policy) for their ipc.

However, it is also not exactly clear where such integration policy would be. It is not part of qm, and it is not part of vsomeip, it is just something related to integrating qm and vsomeip. Clearly this is an area that needs further work, and @mkemel s prototyping in https://gitlab.com/mengel1/custom-qm-policy seems like a good start. Long term I think we need something like that exposed in automotive-image-builder.

Additionally, I *strongly* recommend that we do not use the /run/ipc mechanism for anything but simple demos. It was added as a quick way to "fix" the issue related to IPC between the host and QM, but it is not a good solution for more complex use. The ipc_t type it introduces is used for *all* ipc objects in the qm container which means all security limits for it are shared. To simplify a description of this, suppose we grant qm access two two sockets, but one should only be readable and one only writable. Both sockets will be labeled ipc_t, so the rules will be "qm_t can read ipc_t" and "qm_t can write ipc_t". The existance of these two rules means that qm can (incorrectly) write to the read-only socket, and there is no way around this.

The proper approach is to give each resource that we add to the qm container its own type, to allow rules specific to that resource only. Of course, this ends up being more complicated, but it is something we eventually will have to solve.

shetze:
I opened a [PR for the sample-images](https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/587) that demonstrates the current IPC mechanism. This sample depends on the additional policies added to this PR here.

I agree with @alexlarsson that the IPC mechanism is not what we want for the product. I provide the sample and the additional rm policies merely to share and demonstrate.

Still I would like to see the initial PR to be merged because the vsomeip specific policies do not belong into the qm package.

dougsland:
> I opened a [PR for the sample-images](https://gitlab.com/CentOS/automotive/sample
>-images/-/merge_requests/587) that demonstrates the current IPC mechanism. This 
> sample depends on the additional policies added to this PR here.
> 
> I agree with @alexlarsson that the IPC mechanism is not what we want for the product. I provide the sample and the additional rm policies merely to share and demonstrate.
> 
> Still I would like to see the initial PR to be merged because the vsomeip specific policies
> do not belong into the qm package.
>
> I opened a [PR for the sample-images](https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/587) that demonstrates the current IPC mechanism. This sample depends on the additional policies added to this PR here.
> 
> I agree with @alexlarsson that the IPC mechanism is not what we want for the product. I provide the sample and the additional rm policies merely to share and demonstrate.
> 
> Still I would like to see the initial PR to be merged because the vsomeip specific policies do not belong into the qm package.

Yes, we’re all aligned that including VSOMEIP-specific policies in the QM package isn’t the right approach. 💯

Yariv and I explored the use case you described quite a bit a few weeks back and encountered the same SELinux policy issues. At the time, we kept the /run/ipc setup to support existing use cases, demos and keep the SELinux police change as simple as possible to avoid introducing holes.

Also, we were introduced to [custom-qm-policy](https://gitlab.com/mengel1/custom-qm-policy), which looks promising.

That said, I understand the current challenges you're facing with the apps, and your patch seems like a reasonable interim solution.

@Yarboa — WDYT?

PR #907, opened 2025-08-05
subsystem: remove ros2, sound and others

dougsland:
All these subpackages will be replaced by OCI hooks.

## Summary by Sourcery

Enhancements:
- Delete ros2, sound, text2speech, video, and windowmanager subsystems and their container configurations


aesteve-rh (approved):
lgtm

Just to ensure that it is clear, I did not remove these when we included OCI hooks because hooks do not replace quadlets, containerfiles, etc. Those parts of the subpackages do not have a replacement.

That said, I think it is good to remove them and re-think how do we want to ship them.

PR #929, opened 2025-09-19
Added empty qm.containers.d to QM RPM

engelmi:
Added empty qm.containers.d to QM RPM. 
Currently, this drop-in directory needs to be created on the target system each time individually.

## Summary by Sourcery

Automate the creation of the qm.container.d drop-in directory during installation by extending the Makefile and RPM spec to include the new directory, removing the need for manual setup on target systems.

New Features:
- Include an empty qm.container.d drop-in directory in the QM RPM

Enhancements:
- Update the Makefile install target and RPM spec to create the containers/systemd/qm.container.d directory automatically


aesteve-rh (approved):

Yarboa (approved):
LGTM

PR #946, opened 2025-11-10
improve "qm-oom-score-adj": extracting  a "podman inspect" method

pengshanyu:
Enhancements:
extracting "podman inspect" method
replace "echo" to "info_message"
catching stderr in "podman inspect" method, to ensure errors could be caught and monitored

### DIFF
--- a/tests/ffi/qm-oom-score-adj/test.sh
+++ b/tests/ffi/qm-oom-score-adj/test.sh
@@ -20,6 +20,26 @@ trap disk_cleanup EXIT
 prepare_test
 reload_config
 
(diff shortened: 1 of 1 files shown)


Yarboa (commented):
Consider moving this 
`podman inspect "$container_name" --format '{{.State.Pid}}' 2>&1 | tail -1`
To a method, with 2 args, ctr_name format,

It could save the fix in 3 places

Yarboa:
> WARN[0000] The storage 'driver' option should be set in /etc/containers/storage.conf.

when did that happen, the warnings? 
Maybe it is better to understand what is missing

PR #956, opened 2026-01-13
We are attempting to change selinux option which SHOULD create alert

pbrilla-rh:
- adding avc check for this test case
- result: xfail - we want to create avc error and have it logged

## Summary by Sourcery

Tests:
- Introduce an SELinux test case that changes an option to generate an AVC error and verifies it is logged as an expected failure.


Yarboa (commented):
LGTM

Related to libkrun please file an issue

Issue #394, opened 2024-04-26
qm - ostree image cannot start qm service - /etc/localtime is readonly


dougsland:
We need further investigation why podman/quadlet cannot run the container in such scenario. 

- Is it a regression? 
- New bug? 
- As soon we get the fix, revert the workaround https://github.com/containers/qm/pull/393
- The initial report is here: #367 

```
/usr/bin/podman run --name=qm --replace --rm --cgroups=split --tz=local --network=host --sdnotify=conmon -d --security-opt label=type:qm_t --security-opt label=filetype:qm_file_t --security-opt label=level:s0 --device=/dev/fuse --cap-add=all --read-only -v ${RWETCFS}:/etc -v ${RWVARFS}:/var --pids-limit=-1 --security-opt label=nested --security-opt unmask=all --rootfs ${ROOTFS} /sbin/init
Error: removing /etc/localtime: read-only file system 
```

```
podman info  
Error: overlay: can't stat imageStore dir /usr/share/containers/storage: stat /usr/share/containers/storage: no such file or directory 
```

nsednev:
Here is the details of normally running qm container within the `auto-osbuild-rcar_s4-rhivos-fusa-minimal-ostree-aarch64-11433778.29fd8cf6` image, which I ran on aarch64 machine and its ostree is:

```
[root@localhost ~]# ostree --version
libostree:
 Version: '2025.1'
 Git: b028b696b88e668d57f291afa6a70694cfde3e96
 Features:
  - inode64
  - initial-var
  - libcurl
  - libsoup
  - gpgme
  - composefs
  - ex-fsverity
  - libarchive
  - selinux
  - openssl
  - sign-ed25519
  - libmount
  - systemd
  - release
  - p2p
```

An image is running just fine with qm container up and running.

```
[root@localhost ~]# podman ps
CONTAINER ID  IMAGE       COMMAND     CREATED      STATUS      PORTS       NAMES
a2dfd06a7dc1              /sbin/init  3 hours ago  Up 3 hours              qm
[root@localhost ~]# uname -i
aarch64
[root@localhost ~]# uname -a
Linux localhost 5.14.0-570.518.12.1.11.el9_6iv.aarch64 #1 SMP PREEMPT_RT Mon Apr 7 19:13:29 UTC 2025 aarch64 aarch64 aarch64 GNU/Linux
[root@localhost ~]# cat /etc/localtime 
TZif2UTCTZif2UTC
UTC0
[...]

PR #977, opened 2026-07-01
Add User Namespaces

engelmi:
This PR adds the support for running QM with User Namespaces. For this, it 
- adds a systemd sysusers configuration to set up qm-specific users on the host, 
- adds the `SubUIDMap` and `SubGIDMap` to the quadlet, 
- and creates the necessary entries in `/etc/subuid` and `/etc/subgid` during the post-install phase of the RPM install

~It also removes the tz workaround since the issues to fix it are resolved:~
- ~https://github.com/containers/qm/issues/367~
- ~https://github.com/containers/qm/issues/394~

## Summary by Sourcery

Add qm-specific system users and user namespace mappings to support running QM with user namespaces, and clean up obsolete timezone workaround tooling.

Enhancements:
- Introduce a qm-sysusers configuration defining qm-specific users and groups with fixed IDs for user namespace operation.

Build:
- Update RPM spec to install a systemd sysusers configuration for qm-specific users and manage subuid/subgid ranges for qm_root during post-install and uninstall.
- Adjust RPM dependencies to include shadow-utils for user management and tweak SELinux policy-related requirements.

Chores:
- Remove the comment-tz-local script and its RPM packaging references now that the underlying timezone issues are resolved.

## Summary by Sourcery

Add packaging support for QM to run with dedicated user and group namespaces managed via systemd-sysusers and subuid/subgid mappings.

New Features:
- Define QM-specific system users and groups via a systemd sysusers configuration to back user-namespace-based execution.
- Provision subuid and subgid ranges for the qm_root user during RPM post-install to enable user namespaces.

Enhancements:
- Update the RPM spec to bump the package version and ensure required tools (e.g. shadow-utils) are available for user management.

### DIFF
--- a/rpm/qm.spec
+++ b/rpm/qm.spec
@@ -64,12 +64,13 @@ BuildRequires: selinux-policy-devel >= %_selinux_policy_version
 BuildRequires: bluechi-selinux
 BuildRequires: python3-devel
 
+Recommends: selinux-policy-targeted >= %_selinux_policy_version
+
 Requires: parted
 Requires: containers-common
 Requires: selinux-policy >= %_selinux_policy_version
 Requires(post): selinux-policy-base >= %_selinux_policy_version
 Requires(post): selinux-policy-any >= %_selinux_policy_version
-Recommends: selinux-policy-targeted >= %_selinux_policy_version
 Requires(post): policycoreutils
 Requires(post): libselinux-utils
 Requires: podman >= %{podman_epoch}:4.5
@@ -169,6 +170,7 @@ fi
 %{_datadir}/qm/contexts
 %{_datadir}/qm/file_contexts
 %{_datadir}/qm/setup
+%{_datadir}/qm/userns
 %{_datadir}/qm/create-seccomp-rules
 %{_datadir}/qm/qm-rootfs
 %{_datadir}/qm/qm-storage-settings
--- a/setup
+++ b/setup
@@ -18,8 +18,6 @@ RWETCFS="/etc/qm"
 RWVARFS="/var/qm"
 AGENT_HOSTNAME="$(hostname)"
 AGENTCONF="/etc/bluechi/agent.conf.d/agent.conf"
-QM_CONTAINER_IDS=1000000000:1500000000
-CONTAINER_IDS=2500000000:1500000000
 PACKAGES_TO_INSTALL="selinux-policy-targeted podman systemd procps-ng iptables-nft jq hostname"
 
 # RHEL kernel uses iptables-nft, not iptables-legacy, the tool should
@@ -99,11 +97,6 @@ usage()
    echo "  $ sudo ./setup --installdir=/usr/share/qm --rootfs=/usr/lib/qm/rootfs"
 }
 
-replaceIDs() {
-    touch "$1"
-    grep -q "^$2:" "$1" || echo "$2":"$3" >> "$1"
-}
-
 bluechiSetup() {
     ROOTFS=$1
     RWETCFS=$2
@@ -131,8 +124,10 @@ setupRW() {
     ROOTFS=$1
     RWETCFS=$2
     RWVARFS=$3
+
     mkdir -Z -p "${ROOTFS}/etc" "${ROOTFS}/var"
     mkdir -Z -p "${RWETCFS}" "${RWVARFS}"
+
     mount --bind "${RWETCFS}" "${ROOTFS}/etc"
     mount --bind "${RWVARFS}" "${ROOTFS}/var"
 }
@@ -263,8 +258,6 @@ EOF
     cp "${INSTALLDIR}/containers.conf" "${ROOTFS}/etc/containers/"
     cp "${INSTALLDIR}/contexts" "${ROOTFS}/usr/share/containers/selinux/"
     cp "${INSTALLDIR}/file_contexts" "${ROOTFS}/etc/selinux/targeted/contexts/files/file_contexts"
-    replaceIDs "${ROOTFS}/etc/subuid" containers ${QM_CONTAINER_IDS}
-    replaceIDs "${ROOTFS}/etc/subgid" containers ${QM_CONTAINER_IDS}
     bluechiSetup "${ROOTFS}" "${RWETCFS}" "${RWVARFS}"
 
     if [ "$SYSTEMCTL_SKIP" == "N" ]; then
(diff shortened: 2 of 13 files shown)


rpm/qm.spec
```diff
@@ -127,15 +128,20 @@ install -m 0644 tools/qmctl/qmctl.py %{buildroot}%{python3_sitelib}/qmctl/
...
 . %{_sysconfdir}/selinux/config
 %selinux_modules_install -s ${SELINUXTYPE} $MODULES
 # Execute the script to create seccomp rules after the package is installed
 /usr/share/qm/create-seccomp-rules
-/usr/share/qm/comment-tz-local # FIX-ME GH-issue: 367
 # podmand netavark requires at host to load or let's ignore in case host don't have it and proceed with the installation
 modprobe ip_tables || true
+# Add user namespace entries to /etc/subuid and /etc/subgid for the qm_root user
+usermod --add-subuids 1000000000-2500000000 qm_root
+usermod --add-subgids 1000000000-2500000000 qm_root
```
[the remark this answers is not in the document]
engelmi: This is fixed by triggering explicitly systemd-sysusers before the usermod.

rpm/qm.spec
```diff
@@ -153,6 +159,9 @@ if [ $1 -eq 0 ]; then
    if [ -f %{seccomp_json} ]; then
      /bin/rm -f %{seccomp_json}
    fi
+   # Remove user namespace entries from /etc/subuid and /etc/subgid for the qm user
+   usermod --del-subuids 1000000000-2500000000 qm_root  &> /dev/null
+   usermod --del-subgids 1000000000-2500000000 qm_root  &> /dev/null
```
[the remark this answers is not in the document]
engelmi: Changed. Also added an upfront if check

qm-sysusers.conf
```diff
@@ -0,0 +1,57 @@
...
+u       qm_operator          1000000011:1000000000      "QM - operator"                    /root                 /sbin/nologin
+u       qm_games             1000000012:1000000020      "QM - games"                       /usr/games            /sbin/nologin
+u       qm_ftp               1000000014:1000000050      "QM - FTP User"                    /var/ftp              /sbin/nologin
+u       qm_tss               1000000059:1000000059      "QM - Account used for TPM access" /                     /sbin/nologin
+u       qm_sshd              1000000074:1000000074      "QM - Privilege-separated SSH"     /usr/share/empty.sshd /usr/sbin/nologin
+u       qm_dbus              1000000081:1000000081      "QM - System message bus"          /                     /sbin/nologin
+u       qm_chrony            1000000998                 "QM - chrony system user"          /var/lib/chrony       /sbin/nologin
+u       qm_systemd-coredump  1000000999                 "QM - systemd Core Dumper"         /                     /sbin/nologin
+u       qm_nobody            1000065534                 "QM - Kernel Overflow User"        /                     /sbin/nologin
```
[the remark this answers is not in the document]
engelmi: Added the explicit gid mapping.

engelmi:
After doing some builds with aib, I figured out that the tz workaround seems to be still needed - so I dropped that commit from this PR.

rpm/qm.spec
```diff
@@ -137,6 +141,18 @@ install -m 0644 tools/qmctl/qmctl.py %{buildroot}%{python3_sitelib}/qmctl/
...
 modprobe ip_tables || true
 
+# Trigger systemd-sysusers to create all QM users now
+# Makes the users available for the subsequent usermod
+systemd-sysusers
+# Add user namespace entries to /etc/subuid and /etc/subgid for the qm_root user
+#
+if ! grep -q "^qm_root:" /etc/subuid 2>/dev/null; then
+   usermod --add-subuids 1000000000-2500000000 qm_root
+fi
```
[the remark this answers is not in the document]
engelmi: Without the proper subuids/subgids QM would fail anyway. However, its probably easier to just edit the /etc/sub[u|g]id files anyway. Changing.

qm-sysusers.conf
```diff
@@ -0,0 +1,57 @@
...
+g       qm_users             1000000100
+g       qm_ssh_keys          1000000101
+g       qm_clock             1000000103
+g       qm_systemd-journal   1000000190
+g       qm_chrony            1000000995
+g       qm_sgx               1000000996
+g       qm_systemd-coredump  1000000997
+g       qm_render            1000000998
+g       qm_input             1000000999
+g       qm_nobody            1000065534
```
alexlarsson: While I agree with this idea, i wonder if its correct to hardcode this list in the general qm package. I mean, it is specific to the cs10 user list. Maybe it could be used as a base, and then we replace it in a-i-b with a generated one based on the actual qm users?

alexlarsson: Maybe what we want is to add some tool to qm to generate this, and then we'd have a-i-b launch that at build time?

engelmi: Yes, a tool for generating the sysusers conf makes sense. There is already the [setup script](https://github.com/containers/qm/blob/main/setup) shipped with QM - to bootstrap QM in non-aib environments. So we could do something similar and then invoke it via a custom osbuild-auto stage `org.osbuild.qm.userns` in aib.

It should be 
- able to get the users from inside of QM (which should always be `/etc/qm/passwd` with [current quadlet configuration](https://github.com/containers/qm/blob/main/qm.container#L122))
- add the subuid/subgid to it as an offset 
- and save it as a sysusers config in the root partition.

engelmi: It probably makes sense to move the `usermod` for the subuid/subgid assignment into that script as well - then its also not hard-coded.

engelmi: Decided to completely isolate the userns support in the new tool (the `userns` bash script). This way we do not break anything, can use and refine it. And when we are certain nothing breaks, we can make it the default.

Yarboa (approved):
LGTM
Opened ticket for CI failures
I would expect userns integration tests other then unittest

engelmi:
> LGTM
> Opened ticket for CI failures 
> I would expect userns integration tests other then unittest

Thanks. 
The initial intend was to simply run all existing tests with user namespaces enabled (which would be the integration tests you mentioned). However, since the tests are currently failing, adding another component just complicates it. So fixing them is - sort of - a blocker for the integration tests. There are additional complexities (e.g. the mounted rootfs being owned by root etc.), so I'll take care of this after the tests are fixed in a separate PR.
And unit tests are cheap to run etc., so its good to have them for this in place :)

setup
```diff
@@ -18,8 +18,6 @@ RWETCFS="/etc/qm"
 RWVARFS="/var/qm"
 AGENT_HOSTNAME="$(hostname)"
 AGENTCONF="/etc/bluechi/agent.conf.d/agent.conf"
-QM_CONTAINER_IDS=1000000000:1500000000
```
dougsland, line 21: @engelmi do we need to remove these lines in setup script too as this is optional?

dougsland:
LGTM but have small question

setup
engelmi: These ID ranges are added for the qmcontainers and containers user, which were and are not used. So its - sort of - dead code. I mapped the range here for the default values in `userns` - which would cause collisions.

PR #984, opened 2026-09-08
qm.spec: rename qm-ctl to qmctl

dougsland:
The tool is named qmctl, so update the spec file to match the tool's name.

## Summary by Sourcery

Enhancements:
- Rename the RPM controller subpackage to `qmctl` to match the tool name while preserving compatibility with the former `qm-ctl` package.


Yarboa (approved):
LGTM

Issue #986, opened 2026-09-09
TestingFarm: Attempting to create queue /test_queue_0... FAILED: Permission denied


dougsland:
CentOS 9 - CI/CD is failing 

```
Complete!
[ INFO  ] PASS: Command 'dnf install --setopt=reposdir=/etc/yum.repos.d  --installroot=/usr/lib/qm/rootfs -y gcc make || true' successful
[ INFO  ] PASS: Command 'podman exec qm mkdir -p /tmp/mqueue_test' successful
[ INFO  ] PASS: Command 'podman cp test_mqueue_limit.c qm:/tmp/mqueue_test/' successful
[ INFO  ] PASS: Command 'podman cp Makefile qm:/tmp/mqueue_test/' successful
gcc -Wall -Wextra -std=c99 -o test_mqueue_limit test_mqueue_limit.c -lrt
[ INFO  ] PASS: Command 'podman exec qm bash -c 'cd /tmp/mqueue_test && make test_mqueue_limit'' successful
Testing message queue limit enforcement...
Attempting to create queue /test_queue_0... FAILED: Permission denied
Unexpected error: Permission denied
[ FAILED ] Error executing command 'podman exec qm bash -c 'cd /tmp/mqueue_test && ./test_mqueue_limit'' with exit code: 1
[ INFO  ] Running comprehensive cleanup
[ INFO  ] Cleaning up leftover host message queue: /dev/mqueue/host_test_queue_before_1
[ INFO  ] Cleaning up leftover host message queue: /dev/mqueue/host_test_queue_before_2
[ INFO  ] Cleaning up leftover host message queue: /dev/mqueue/host_test_queue_before_3
[ INFO  ] Cleaning up leftover host message queue: /dev/mqueue/host_test_queue_before_4
[ INFO  ] Cleaning up leftover host message queue: /dev/mqueue/host_test_queue_before_5
[ INFO  ] Cleaning up container message queues
[ INFO  ] Cleaning up host compiled test program
rm -f test_mqueue_limit test_host_mqueue
8d2766365e5f6ecd1fad782d46b7f3dca8fcd71d280e5ac9853141ca09c46baf
[ INFO  ] PASS: Command 'podman exec qm /usr/bin/podman rm -af' successful
[ INFO  ] PASS: Command 'rm -rf /etc/containers/systemd/qm.container.d/' successful
[ INFO  ] PASS: Command 'systemctl daemon-reload' successful
[...]

PR #991, opened 2026-09-17
Qm 1.1.z

dougsland:
## Summary by Sourcery

Prepare the 1.1.z release with the renamed `qmctl` RPM package and backward-compatible packaging metadata.

Enhancements:
- Rename the RPM controller subpackage to `qmctl` while preserving compatibility with the previous package name and ensuring versioned dependencies include the package epoch.

Build:
- Update the package specification for the renamed `qmctl` RPM subpackage.

Chores:
- Bump the project version and SELinux policy version metadata.


dougsland:
There are issues already opened to address these ci/cd errors. Not to related to the patch.
We also need to deprecate the centos-stream-9.
https://github.com/containers/qm/issues/986

